PR #1, opened 2019-06-14
Switch the configuration from Dahll to Yaml

ndmitchell:
Note that this patch is completely untested (other than it builds), since you didn't seem to have any Dahll test files sitting around. I went for a pretty simple encoding, so `cabal:` or `cabal: {component: lib}` are sufficient. It probably doesn't produce fantastic error messages.

### DIFF
--- a/src/HIE/Bios/Config.hs
+++ b/src/HIE/Bios/Config.hs
@@ -1,13 +1,14 @@
-{-# LANGUAGE DeriveGeneric     #-}
 {-# LANGUAGE OverloadedStrings #-}
-{-# LANGUAGE TypeApplications #-}
-module HIE.Bios.Config where
+{-# LANGUAGE ViewPatterns #-}
+module HIE.Bios.Config(
+    readConfig,
+    Config(..),
+    CradleConfig(..)
+    ) where
 
-import Dhall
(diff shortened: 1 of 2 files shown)


src/HIE/Bios/Config.hs
```diff
@@ -16,33 +17,25 @@ data CradleConfig = Cabal { component :: Maybe String }
...
-                  deriving (Generic, Show)
-
-instance Interpret CradleConfig
+                  deriving (Show)
+
+instance FromJSON CradleConfig where
+    parseJSON (Object (Map.toList -> [(key, val)]))
+        | key == "cabal" = case val of
+            Object x | Just (String v) <- Map.lookup "component" x -> return $ Cabal $ Just $ T.unpack v
+            _ -> return $ Cabal Nothing
```
mpickering: This is going to be confusing because it will accept `cabal: garbage`. Something to change in future.

mpickering:
Good start. I'll merge this in but needs to be modified before being released (imo).

mpickering:
Also, all the documentation needs to be updated.

ndmitchell:
Thanks - agreed - more testing would be good too.

PR #7, opened 2019-06-16
Make cabal cradle work on Windows

ndmitchell:
Also renames biosc to hie-bios.


mpickering:
Thanks a lot @ndmitchell

PR #8, opened 2019-06-16
Add a Direct cradle

ndmitchell:
Useful for just listing out the command lines - I've used this to get it working for Shake


mpickering:
Thanks, great idea.

PR #15, opened 2019-08-16
Fix stack cradle

cocreature:
Both the stack and the cabal cradle use the same wrapper script which
outputs the working directory as the first argument to fix import
paths. However, the stack cradle was never post processing the
arguments which resulted in the working directory being passed to GHC
which complains that it’s not a module.


isovector:
I literally just ran into this issue; this fixes it for me too. stack lts-14.0

mpickering:
Thanks, I need to add some tests! 🙅‍♂

PR #17, opened 2019-08-31
Load plugins

### LINKED ISSUE #16: hie-bios doesn't enable GHC plugins
`loadFile` will fail on modules that require a GHC plugin to compile. It doesn't seem to respect `-fplugin` options set in either the cradle or via `OPTIONS_GHC`.

isovector:
Fixes #16

### DIFF
--- a/src/HIE/Bios/Load.hs
+++ b/src/HIE/Bios/Load.hs
@@ -11,7 +11,12 @@ import HscTypes
 import Outputable
 import Control.Monad.IO.Class
 
+#if __GLASGOW_HASKELL__ >= 806
+import DynamicLoading
+#endif
+
 import Data.IORef
+import Data.Traversable
 
 import System.Directory
 import Hooks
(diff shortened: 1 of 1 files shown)


isovector:
@mpickering you suggested on IRC that this might not be the right solution. It's pretty slow on my end, wondering if the speed is due to this not being right.

mpickering:
How slow are we talking here? I don't think my idea would make it any faster.

isovector:
We're talking my editor is unusable on a moderately sized project with ~15 modules. My guess is that `load` isn't caching unchanged modules.

isovector:
(but i haven't dug into the perf; was just sorta hoping you'd divine the right answer :) )

mpickering:
You need to pass the right flags to make it cache interface files. Are you passing `-fwrite-interface -hidir=...`?

mpickering:
You are also supposed to cache the result of calling `findCradle`. If you don't do that it will be very slow

mpickering:
I'm going to look into this after #53 is merged.

mpickering:
I tried out the current master with a simple plugin which just calls error. By invoking `hie-bios` using the `check` command it simulates an initialisation followed by a simple call to `loadFile`. 

In both cases, plugin enabled in module and plugin enabled in cabal file, the load certainly invoked the plugin. 

Can you please provide a sample project that we can try to reproduce this issue?

mpickering:
I put up a more refined version of this patch in #74

PR #18, opened 2019-09-06
Correct extension of configuration file examples

jneira:
Afaiu the package don't use dhall directly to read the configuration.
Really it would be great that dhall would be supported though! :smile:


mpickering:
Thanks @jneira. The config file did used to be dhall but it was too spicy for @ndmitchell so we changed it.

ndmitchell:
🌶

Issue #20, opened 2019-09-06
Unexpected colon in cradle example.


philderbeast:
From the readme, trying the `cradle: {stack:}` example for `hie.yaml` I get an error with `hie-core` and `yamllint`.

https://github.com/mpickering/hie-bios/blame/master/README.md#L38

```
> hie-core
Starting hie-core
...
[1/6] Finding hie-bios cradle
hie-core: InvalidYaml (Just (YamlParseException {yamlProblem = "found unexpected ':'"
, yamlContext = "while scanning a plain scalar"
, yamlProblemMark = YamlMark {yamlIndex = 14, yamlLine = 0, yamlColumn = 14}}))
```

Trying [yamllint](http://www.yamllint.com/)

```
> cradle: {stack:}
found unexpected ':' while scanning a plain scalar at line 1 column 10
```

mpickering:
Thanks, fixed.

PR #21, opened 2019-09-06
Remove troublesome colon in hie.yaml example.

philderbeast:
See #20.


philderbeast:
The linter OKs the fix and reformats it as:

```yaml
--- 
cradle: 
  stack: ~
```

mpickering:
I don't understand yaml it seems! Thanks for making the fix.

PR #22, opened 2019-09-06
A couple of fixes needed to use hie-core on Darcs

hsenag:
Darcs is big enough that the ghc command-line is about 11K characters which blows the limit for
Windows batch files. We discussed this a bit in person at Munihac and decided this was the least
invasive solution.

Also needed to lift a bit more code from ghc's Main.hs to support recognising '.o' arguments coming
from C source code and passing them to the right place.

### DIFF
--- a/src/HIE/Bios/Cradle.hs
+++ b/src/HIE/Bios/Cradle.hs
@@ -138,14 +138,19 @@ processCabalWrapperArgs args =
             in trace dir $ Just final_args
         _ -> Nothing
 
-cabalAction :: FilePath -> Maybe String -> FilePath -> IO (ExitCode, String, [String])
-cabalAction work_dir mc _fp = do
+getCabalWrapperTool :: IO FilePath
+getCabalWrapperTool = do
   wrapper_fp <- writeSystemTempFile "wrapper.bat" $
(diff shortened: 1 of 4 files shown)


src/HIE/Bios/Cradle.hs
```diff
@@ -138,14 +138,29 @@ processCabalWrapperArgs args =
...
+  wrapper_fp <-
+    if isWindows
+      then do
+        wrapper_hs <- writeSystemTempFile "wrapper.hs" cabalWrapperHs
+        -- the initial contents will be overwritten immediately after by ghc
+        wrapper_fp <- writeSystemTempFile "wrapper.exe" ""
+        callProcess "ghc" ["-o", wrapper_fp, wrapper_hs]
+        return wrapper_fp
+      else do
+        writeSystemTempFile "wrapper.bat" cabalWrapper
```
mpickering: We probably shouldn't call it `wrapper.bat` if it's only used on linux?

hsenag: Agreed, I was keeping the old code in the else branch but it seems like an obvious cleanup to change it back.

PR #24, opened 2019-09-07
Run ghc for the wrapper in the temp directory to avoid .ghc-environment

hsenag:


src/HIE/Bios/Cradle.hs
```diff
@@ -149,7 +149,9 @@ getCabalWrapperTool = do
         wrapper_hs <- writeSystemTempFile "wrapper.hs" cabalWrapperHs
         -- the initial contents will be overwritten immediately after by ghc
         wrapper_fp <- writeSystemTempFile "wrapper.exe" ""
-        callProcess "ghc" ["-o", wrapper_fp, wrapper_hs]
+        let ghc = (proc "ghc" ["-o", wrapper_fp, wrapper_hs])
+                    { cwd = Just (takeDirectory wrapper_hs) }
+        readCreateProcess ghc "" >>= putStr
```
mpickering: Is this `putStr` for debugging?

hsenag: It's to replicate the previous behaviour of `callProcess` which just passed through stdout directly. There doesn't seem to be a `callProcess`  equivalent if you need a custom `CreateProcess` value.

mpickering: You should use `createProcess` rather than `readCreateProcess` then? `proc` by default inherits the stdout handle.

hsenag: That would work, but then I'd have to copy and paste the bits of callProcess that throw an error if the command fails (or we decide we don't care about that).

PR #31, opened 2019-09-08
Drop Alternative instance for Ghc

fendor:


mpickering:
Assuming this builds, looks good. I should add some CI.

PR #34, opened 2019-09-12
Increase Haskell bounds for 8.8

ocharles:
This library builds fine on GHC 8.8.1 with jailbroken bounds. This commit increases the bounds to the latest bounds on Hackage.


mpickering:
Thanks Ollie.

PR #39, opened 2019-09-13
WIP: Add dependencies of cradles

### LINKED ISSUE #35: Report what files are dependencies of the Cradle
See https://github.com/digital-asset/ghcide/issues/50, which wants ghcide to reload when you update your .cabal file. One option is that a `Cradle` should have an extra field `cradeDependencies :: [FilePath]` which would give the files which the cradle uses.

fendor:
closes #35
Adds a new record-field for files that are dependencies of the cradle.
cc @ndmitchell

### DIFF
--- a/src/HIE/Bios/Cradle.hs
+++ b/src/HIE/Bios/Cradle.hs
@@ -109,7 +109,7 @@ directCradle :: FilePath -> [String] -> Cradle
 directCradle wdir args =
   Cradle {
       cradleRootDir = wdir
-    , cradleDependencies = ["hie.yaml"] -- ^ Flags are supplied via 'hie.yaml'
+    , cradleDependencies = [] -- ^ Flags are supplied via 'hie.yaml'
     , cradleOptsProg = CradleAction "direct" (const $ return (ExitSuccess, "", args))
   }
 
--- a/src/HIE/Bios/Cradle.hs
+++ b/src/HIE/Bios/Cradle.hs
@@ -41,9 +41,9 @@ loadImplicitCradle :: FilePath -> IO Cradle
 loadImplicitCradle wfile = do
   let wdir = takeDirectory wfile
   cfg <- runMaybeT (implicitConfig wdir)
-  return $ case cfg of
+  case cfg of
     Just bc -> getCradle bc
-    Nothing -> defaultCradle wdir
+    Nothing -> return $ defaultCradle wdir
 
 -- | Finding 'Cradle'.
 --   Find a cabal file by tracing ancestor directories.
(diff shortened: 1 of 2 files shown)


src/HIE/Bios/Cradle.hs
```diff
@@ -107,6 +109,7 @@ directCradle :: FilePath -> [String] -> Cradle
 directCradle wdir args =
   Cradle {
       cradleRootDir = wdir
+    , cradleDependencies = ["hie.yaml"] -- ^ Flags are supplied via 'hie.yaml'
```
fendor: Actually, every cradle might need to monitor "hie.yaml" if it exists, since it can change the used cradle.

src/HIE/Bios/Cradle.hs
```diff
@@ -138,13 +142,20 @@ biosAction _wdir bios fp = do
...
 
-cabalCradle :: FilePath -> Maybe String -> Cradle
-cabalCradle wdir mc =
-  Cradle {
+cabalCradle :: FilePath -> Maybe String -> IO Cradle
+cabalCradle wdir mc = do
+  cabalFiles <- findCabalFiles wdir
+  return Cradle {
       cradleRootDir    = wdir
+    , cradleDependencies = ["cabal.project", "cabal.project.local"] ++ cabalFiles
```
fendor: are there any overlooked special cases?

src/HIE/Bios/Cradle.hs
```diff
@@ -267,6 +280,7 @@ rulesHaskellCradle :: FilePath -> Cradle
 rulesHaskellCradle wdir =
   Cradle {
       cradleRootDir  = wdir
+    , cradleDependencies = ["BUILD.bazel", "WORKSPACE"]
```
fendor: I dont use rules_haskell, so I am not sure if these are all cradle dependencies

src/HIE/Bios/Cradle.hs
ndmitchell: I don't think we need to put hie.yaml on the list - that will always go on the list automatically since all of them use it.

src/HIE/Bios/Types.hs
```diff
@@ -146,12 +146,13 @@ quote opt str = ("\"" ++) .  (quote' str ++) . ("\"" ++)
...
-data Cradle = Cradle {
-  -- | The project root directory.
-    cradleRootDir    :: FilePath
-  -- | The action which needs to be executed to get the correct
-  -- command line arguments
-  , cradleOptsProg   :: CradleAction
+data Cradle = Cradle
+  { cradleRootDir    :: FilePath -- ^  The project root directory.
+  , cradleDependencies :: [FilePath] -- ^ Dependencies of a cradle.
+  -- E.g. for cabal, '*.cabal' files, for stack "stack.yaml"
```
ndmitchell: I'd add a note that some of the files may not exist.

fendor: Thanks for catching it! I also documented that the FilePaths are relative to the cradleRootDir.

src/HIE/Bios/Cradle.hs
fendor: This will not work if the cabal file is in some sub-directory.
We could recursively search all sub-directories for cabal files, or parse "cabal.project" and "stack.yaml" respectively to get the subdirectories we are interested in.

ndmitchell: It's close. Ideally stack and cabal would spit this information out themselves.

fendor:
Another idea that works around most of the problems: Add a field to hie.yaml "cradle-deps" where a user can specify the deps for a cradle.

fendor:
@mpickering suggested that it should be possible to query the build tool for cradle dependencies, e.g. then the signature would be `cradleDependencies :: IO [FilePath]`.

ndmitchell:
Would it be cheaper to query the build tool for cradle dependencies simultaneously with querying it for flags? In which case an extra field in the return of `CradleAction` might work best.

fendor:
Closed in favour of #42

PR #42, opened 2019-09-16
Add searchable cradle dependencies

### LINKED ISSUE #35: Report what files are dependencies of the Cradle

fendor:
Extends hie.yaml file format with the "dependencies" object.
Extends Bios cradle type with an optional FilePath to
obtain cradle dependencies from.

Set for each cradle type the cradle dependencies.

Reorganise parsing of "hie.yaml".

* [x] For each cradle type, implement a function that returns dependencies common to the respective cradle type.
* [x] Test the parsing of `hie.yaml` a bit more, smoke testing succeeded

Note: parsing of hie.yaml is less strict now, if there is a field such as "dependecies" (note missing 'n'), it will parse, assuming the correct field "dependencies" is not defined. 

Closes #35
Makes #39 obsolete

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,10 +78,11 @@ by the `cabal` cradle.
 cradle:
   cabal:
     component: "lib:hie-bios"
-  dependencies:
-    - package.yaml
-    - shell.nix
-    - default.nix
+
+dependencies:
+  - package.yaml
+  - shell.nix
+  - default.nix
 ```
 
 For the `Bios` cradle type, there is an optional field to specify a program
@@ -116,8 +117,8 @@ cradle:
     arguments: ["list","of","ghc","arguments"]
   default:
 
-  dependencies:
-    - someDep
+dependencies:
+  - someDep
 ```
 
 ## Implicit Configuration
--- a/README.md
+++ b/README.md
@@ -53,17 +53,71 @@ relative to the current working directory if it is not an absolute path.
 cradle: {bios: {program: ".hie-bios"}}
 ```
 
+### Cradle Dependencies
+
+Sometimes it is necessary to reload a cradle, for example when a package
+dependency is added to the project. Each type of cradle defines a list of
+files that might cause an existing cradle to no longer provide accurate
+diagnostics if changed. These are expected to be relative to the root of
+the cradle.
+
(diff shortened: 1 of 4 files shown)


src/HIE/Bios/Cradle.hs
```diff
@@ -300,12 +355,18 @@ obeliskWorkDir fp = do
...
   return wdir
 
-
-obeliskCradle :: FilePath -> Cradle
-obeliskCradle wdir =
-  Cradle {
-      cradleRootDir  = wdir
-    , cradleOptsProg = CradleAction "obelisk" (obeliskAction wdir)
+obeliskCradleDependencies :: FilePath -> IO [FilePath]
+obeliskCradleDependencies _wdir = return []
```
fendor: Couldnt find any obelisk files in an very shallow search (mainly the obelisk gh page)

mpickering:
I think just needs a update to the `README`.

README.md
```diff
@@ -53,17 +53,71 @@ relative to the current working directory if it is not an absolute path.
...
+project is compiled.
+
+As an example how you would add cradle dependencies that may not be covered
+by the `cabal` cradle.
+
+```yaml
+cradle:
+  cabal:
+    component: "lib:hie-bios"
+  dependencies:
```
mpickering: Shouldn't the `dependencies` field be at the same level as `cradle`?

mpickering:
Surely the `dependencies` field should be at the top level, like `cradle`? What's your reasoning for the way you have implemented it?

fendor:
> 
> 
> Surely the `dependencies` field should be at the top level, like `cradle`? What's your reasoning for the way you have implemented it?

My reasoning was these are the depedencies of a cradle and a cradle specification consists of a type of cradle and dependencies to that cradle.
But no technical reasoning.

mpickering:
Can you change the `dependencies` to be on the top level of the yaml file?

Then I will merge it. The `cradle` option is really specifying the `CradleType`.

PR #43, opened 2019-09-16
Delimit GHC arguments

ttuegel:
The Cabal wrappers break if there are spaces in any argument to GHC. That can happen if paths have spaces, or with options like `-with-rtsopts=` which expect an argument containing spaces. This would delimit the list of arguments with NULL characters. NULL-terminated arguments allows GHC options to contain spaces without the overhead of Read/Show.

### DIFF
--- a/src/HIE/Bios/Cradle.hs
+++ b/src/HIE/Bios/Cradle.hs
@@ -155,9 +155,20 @@ processCabalWrapperArgs :: String -> Maybe [String]
 processCabalWrapperArgs args =
     case lines args of
         [dir, ghc_args] ->
-            let final_args = removeInteractive $ map (fixImportDirs dir) (words ghc_args)
+            let final_args =
+                    removeInteractive
+                    $ map (fixImportDirs dir)
+                    $ limited ghc_args
             in Just final_args
         _ -> Nothing
+  where
+    limited :: String -> [String]
+    limited = unfoldr $ \argstr ->
+        if null argstr
+        then Nothing
+        else
+            let (arg, argstr') = break (== '\NUL') argstr
+            in Just (arg, drop 1 argstr')
 
 -- generate a fake GHC that can be passed to cabal
 -- when run with --interactive, it will print out its
--- a/wrappers/cabal.hs
+++ b/wrappers/cabal.hs
@@ -17,4 +17,4 @@ main = do
       exitWith code
 
 delimited :: [String] -> String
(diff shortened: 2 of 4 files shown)


src/HIE/Bios/Cradle.hs
```diff
@@ -155,9 +155,20 @@ processCabalWrapperArgs :: String -> Maybe [String]
...
-            let final_args = removeInteractive $ map (fixImportDirs dir) (words ghc_args)
+            let final_args =
+                    removeInteractive
+                    $ map (fixImportDirs dir)
+                    $ limited ghc_args
             in Just final_args
         _ -> Nothing
+  where
+    limited :: String -> [String]
+    limited = unfoldr $ \argstr ->
```
mpickering: Why did you call this `limited`?

wrappers/cabal.hs
```diff
@@ -10,9 +10,11 @@ main = do
...
-      -- note this probably breaks if paths have spaces in
-      putStrLn $ unwords args
+      putStrLn $ delimited args
     _ -> do
       ph <- spawnProcess "ghc" args
       code <- waitForProcess ph
       exitWith code
+
+delimited :: [String] -> String
+delimited = (++) "-here " . foldr (\a as -> a ++ '\NUL' : as) ""
```
ndmitchell: The `foldr` bit is just `intercalate` in disguise.

mpickering: What is this `-here` for?

ttuegel: It's different from `intercalate` because it puts `'\NUL'` characters *after* every list element, instead of *between* elements. `delimited` will put a `'\NUL'` at the end of the list, `intercalate` will not.

ttuegel: That is a piece of debugging output which I have accidentally left in my pull request. :sweat_smile: I will fix that right away!

src/HIE/Bios/Cradle.hs
ttuegel: I didn't think of a better name for the opposite of `delimited`. Of course, I have no attachment to the name and will gladly change it. :smiley:

wrappers/cabal.hs
ndmitchell: `concatMap (++ "\NUL")` ? I find the foldr very hard to interpret, hence my desire to drop it!

PR #51, opened 2019-09-21
Implement the none cradle

mpickering:
The none cradle is meant to be used to tell your IDE to not even bother
trying to load a certain subdirectory.

For example, if your project contains lots of haskell files for testing
then you don't want to try to load these into your IDE when you open
them. If you're working on a project like GHC, many of them will fail to
load or even parse.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,7 @@ cradle:
   direct:
     arguments: ["list","of","ghc","arguments"]
   default:
+  none:
 
 dependencies:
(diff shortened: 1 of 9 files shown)


fendor:
I really like the API changes!

ndmitchell:
API looks good to me - explicit types over Either, and a new Maybe smuggled inside - can't argue with better types and names!

PR #53, opened 2019-09-21
Multi cradle

mpickering:
This patch isn't ready yet as it shows the API is slightly wrong in a few places. Specifically the dependencies API is not quite right, the `CradleAction` should also return the dependencies rather than being a separate function I believe so it can also depend on the `FilePath` as the dependencies may be totally different per component. 

Example config.

```
cradle:                                                                         
   multi:                                                                        
    - path: "./"                                                                
       config: { cradle: {cabal: {component: "lib:hie-bios"}} }  
```

Also it will be useless given the current behaviour of `ghcide` which just always passes the empty pass to the cradle. When it starts passing a more nuanced path and supports multiple components then this configuration may be more useful.

### DIFF
--- a/src/HIE/Bios/Cradle.hs
+++ b/src/HIE/Bios/Cradle.hs
@@ -21,6 +21,7 @@ import Control.Applicative ((<|>))
 import Data.FileEmbed
 import System.IO.Temp
 import Data.List
+import Data.Ord (Down(..))
 
 import System.PosixCompat.Files
 
@@ -168,10 +169,8 @@ multiAction cur_dir cs cur_fp = selectCradle =<< canonicalizeCradles
     -- function we want to choose the most specific cradle possible.
     canonicalizeCradles :: IO [(FilePath, CradleConfig)]
     canonicalizeCradles =
-      sortBy (\(p1, _) (p2, _) ->
-        if p1 `isPrefixOf` p2
-          then GT
-          else LT) <$> mapM (\(p, c) -> (,c) <$> (canonicalizePath (cur_dir </> p))) cs
+      sortOn (Down . fst)
+        <$> mapM (\(p, c) -> (,c) <$> (canonicalizePath (cur_dir </> p))) cs
 
     selectCradle [] =
       return (CradleFail (CradleError ExitSuccess err_msg))
--- a/src/HIE/Bios/Cradle.hs
+++ b/src/HIE/Bios/Cradle.hs
@@ -66,6 +66,7 @@ getCradle (cc, wdir) = addCradleDeps cradleDeps $ case cradleType cc of
     Direct xs -> directCradle wdir xs
     Default   -> defaultCradle wdir
     None      -> noneCradle wdir
+    Multi ms  -> multiCradle wdir ms
     where
       cradleDeps = cradleDependencies cc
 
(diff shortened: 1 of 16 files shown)


ndmitchell:
Seems pretty plausible, although I'm not convinced what the right answer is in this space.

mpickering:
Can you elaborate what you think the right answer is?

One definite non-answer is smashing together lots of components into a single component. It doesn't work in general so it is not a robust solution.

Another non-answer is placing a `hie.yaml` file in the subdirectory of each component. Reason, the `hie.yaml` file has to be placed at the project root so commands operate in the correct context.

ndmitchell:
* I think smashing together lots of components is essential for a good user experience.
* Why not have multiple hie.yaml files and then include a `root` tag on them, to say where the project is rooted? I am currently adding multiple distinct hie.yaml file support to ghcide, and I think it's a very important feature.

mpickering:
Because then you have to litter your build tree with lots of these configuration files. It's also not always the case that a component is contained entirely within one directory. Having a single configuration file which points to each component is easier to understand and specify.

ndmitchell:
Should have been clearer - what you are proposing is not unreasonable, basically we need to know for a given file which HscEnv to use, and which HscEnv's are really the same HscEnv so we can allow cross-imports for them. If this interface can provide that (and it definitely seems like it can) then great.

mpickering:
I think it can @ndmitchell 

In my mind it is just like your solution of specifying multiple `hie.yaml` files but you do so in one file (and give a prefix) rather than committing the files to each subdirectory. 

@fendor Can you have a look at this? I had to modify how the dependencies worked.

mpickering:
I also don't want to merge this until `ghcide` will work properly with multi cradles. It that requires changing the API slightly then I'm happy to do that but I can't easily predict what might need to change.

ndmitchell:
It's really hard to answer the question of what ghcide wants until we all come to a common answer on the question of whether we should try and fudge multiple components into a single GHC Session. I'm going to discuss this with Alan next week.

Currently we make the assumption that each hie.yaml shares the same cradle. That's not essential, and can be changed. What we do need is two methods:

```
fileToState :: FilePath -> IO state
stateToSession :: state -> IO Session
```

The rough rule should be that calling `fileToState` repeatedly doesn't repeatedly call configure etc afresh. If two files share the same state value we should only need to create one session for both of them.

cocreature:
Let me try to summarize my take on multiple components:

1. I think we can all agree that multiple components need to be supported and the default behavior must work out of the box without user intervention (at least in a reasonable number of cases).
2. I am not sure if everyone agrees on this but imho users must be able to choose to load everything as a single component. For Bazel-style monorepos where you have tons of small components, I don’t think loading them separately is a reasonable option. (Note that I’m not saying this must be the default.)
3. For cabal/stack projects (in particular outside of large monorepos which ime are mostly used in commercial development), the number of components is reasonably small so both approach are reasonably and there are arguments for both.
4. Loading everything into a single component has the advantage that reloads are faster and memory usage should be slower. It also matches what you have to do to get `ghci` working across cmoponents so we aren’t really entering new territory here. (That stops being true to some degree once we start making flags per files and only share the package state). The obvious downside is that it doesn’t work if the flags are incompatible. What exactly “incompatible” means here depends on whether we make some flags (like default-extensions) per-file instead of per `HscEnv` or not. I’m leaning towards the former but given that you have to make them compatible anyway for `ghci` to work I don’t think the second option is unreasonable. It also requires us to figure out a somewhat reasonable way of merging flags, maybe the logic in `stack ghci` is a reasonable option.
5. Loading things into multiple `HscEnv`s has the obvious upside that it should work in more cases (I think it’s still not entirely save at least for 8.6 since some state like the RTS linker is global and not per-HscEnv). The downsides correspond to the upsides of using a single component, i.e., slower reloads and slower. It also requires us to implement automatic restarts of components. Adding this to `ghcide` without automatic restarts will lead to lots of confusion so I’m not willing to do that.

Overall, while I think both approaches are reasonable, I prefer loading everything into a single component (as the default, I’m perfectly happy to make this configurable) but I’m happy to be convinced here. In particular, I would be interested in projects where loading things into a single component will fail and isn’t easily fixable (e.g., by adding a package import to disambiguate if you have different deps with colliding module names for two components).

mpickering:
Thanks @cocreature and @ndmitchell, great comments.

I was considering this question a bit this weekend. If we take the simple example of a `cabal` project:

1. The solver will create a consistent plan for the whole project so that each (ie components have to depend on the same version of a package (I believe))
2. If this is the case, then you can use one `HscEnv` for the entire project and compile to module separately and modify the `DynFlags` as appropriate.
3. Therefore we could still assume that one `hie.yaml` meant one `HscEnv`. 

I am still quite nervous this would work correctly so I would like to be convinced by a working implementation! I think my reluctance on this matter has stemmed from limitations of using `ghc --make` where it would certainly not be possible to implement this in a robust manner! It is a big advantage of `ghcide` to reimplement downsweep and make it possible to have more finegrained control over each module's compilation.

ndmitchell:
Sounds like we're reaching agreement - awesome. I think the next step is to write something that can load a cabal project in a custom way with different flags per component etc and prove this approach works in practice. Once we have that proof of concept we can then know what is required to move forward.

mpickering:
@fendor This needs a review please. I split the refactoring of dependencies into a different commit so the introduction of multi cradle is now in its own commit.

It occurred to me that by far the most common way to configure this will be to specify many cabal or stack cradles and in order to align with the assumption of only having one `HscEnv`, that might be the only sensible way to configure your project. Therefore the current implementation might be too general.

src/HIE/Bios/Cradle.hs
```diff
@@ -124,21 +133,63 @@ noneCradle cur_dir =
...
+
+    -- Canonicalize the relative paths present in the multi-cradle and
+    -- also order the paths by most specific first. In the cradle selection
+    -- function we want to choose the most specific cradle possible.
+    canonicalizeCradles :: IO [(FilePath, CradleConfig)]
+    canonicalizeCradles =
+      sortBy (\(p1, _) (p2, _) ->
+        if p1 `isPrefixOf` p2
+          then GT
+          else LT) <$> mapM (\(p, c) -> (,c) <$> (canonicalizePath (cur_dir </> p))) cs
```
fendor: Implementation looks a bit complex, imo.
SortBy could be something like:
`sortOn (Down . compare . fst)` but not a blocker

mpickering: I think it's a good point. I'll change the implementation. I didn't consider that the `Ord` instance for `FilePath` would do the right thing here.

mpickering:
I changed the implementation of `canonicalizeCradles` based on your suggestion.

PR #59, opened 2019-09-25
Cache windows executable wrapper 

### LINKED ISSUE #55: Windows cradles start up slower
Windows Stack/Cabal cradles first compile a binary. It would be much nicer if they didn't have to every time. Perhaps put it in a known location in temp, e.g. hie-bios-windows-helper-v1.exe, and if it already exists avoid recompiling it each time?

jneira:
* Fixes #55
* Using $XDG_CACHE_HOME cache directory as suggested by @mpickering 
* I've used the `hie-bios` version to qualificate the wrapper:
  * We'll have not remember changing its own version if it changes
  * But it will create a 16 Mb executable for each minor version of the library
  * If it doesnt worth it i could use a specific version for it as suggested by @ndmitchell

### DIFF
--- a/src/HIE/Bios/Cradle.hs
+++ b/src/HIE/Bios/Cradle.hs
@@ -247,7 +247,7 @@ getCabalWrapperTool = do
             writeFile wrapper_hs cabalWrapperHs
             createDirectoryIfMissing True cacheDir
             let ghc = (proc "ghc" ["-o", wrapper_fp, wrapper_hs])
-                      { cwd = Just (takeDirectory wrapper_hs) }
+                        { cwd = Just (takeDirectory wrapper_hs) }
             readCreateProcess ghc "" >>= putStr
         return wrapper_fp
       else do
--- a/src/HIE/Bios/Cradle.hs
+++ b/src/HIE/Bios/Cradle.hs
@@ -24,6 +24,8 @@ import Data.List
 
 import System.PosixCompat.Files
 
+import Data.Version (showVersion)
+import Paths_hie_bios
 ----------------------------------------------------------------
 
 -- | Given root/foo/bar.hs, return root/hie.yaml, or wherever the yaml file was found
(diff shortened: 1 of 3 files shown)


ndmitchell:
Using the hie-bios version makes more sense. It's not going to be a huge amount of disk space. This looks great.

src/HIE/Bios/Cradle.hs
```diff
@@ -235,12 +237,16 @@ getCabalWrapperTool = do
...
-        wrapper_fp <- writeSystemTempFile "wrapper.exe" ""
-        let ghc = (proc "ghc" ["-o", wrapper_fp, wrapper_hs])
-                    { cwd = Just (takeDirectory wrapper_hs) }
-        readCreateProcess ghc "" >>= putStr
+        cacheDir <- getXdgDirectory XdgCache "hie-bios"
+        let exeName = ("wrapper-" ++ showVersion version) <.> "exe"
+        let wrapper_fp = cacheDir </> exeName
+        exists <- doesFileExist wrapper_fp
+        unless exists $ do
+            wrapper_hs <- writeSystemTempFile "wrapper.hs" cabalWrapperHs
```
ndmitchell: Why not make it wrapper-version.hs, and avoid creating differently named files for each which are harder to marry up with their exe?

jneira: Mmm i am afraid that `writeSystemTempFile` only uses the first 4 letters of the filename (it creates  `wra4EE1.hs` in my system f.e.) so simply adding the version would not work.
We could create directories in `%TEMP` or save the `.hs` in `%XDG_CACHE_HOME` too (and avoid copying it)

jneira: nvm, we could simply create wrapper-version.hs in TEMP getting ride of `writeSystemTempFile`

jneira: @ndmitchell i think the pr match your proposal, do you think it is ready to be merged?

PR #64, opened 2019-09-27
Change the Cradle to always use the Windows approach of an exe.

myShoggoth:
Change the wrapper code to dedup the GHC arguments, in large
Haskell codebases with many modules and projects we are seeing
the same GHC args repeated a ton (possibly per project? per module?).
Simplifies the approach, allows for more processing to happen on those
args.


mpickering:
I don't really want to merge this. There's no reason to assume that someone has `ghc` on their path. The only reason the windows wrapper works like this is to avoid max path limits.

I also don't understand the problem this attempts to fix: "in large
Haskell codebases with many modules and projects we are seeing
the same GHC args repeated a ton ". How does this occur and why is it a problem?

PR #67, opened 2019-10-03
Implement logging for hie-bios

fendor:
Adds optional dependency for hslogger which is turned off by default.
The logging should not touch stuff for ghcide, cc @ndmitchell.
@mpickering Is this how you would imagine logging?


fendor:
Maybe this branch should rather be a PR against multi-cradle?

mpickering:
Needs to be targeted against the right branch.

PR #69, opened 2019-10-04
Implement logging with optional hslogger dependency

fendor:


hie-bios.cabal
```diff
@@ -18,6 +18,10 @@ Extra-Source-Files:     ChangeLog
 
 tested-with: GHC==8.6.4, GHC==8.4.4, GHC==8.2.2
 
+flag hslogger
```
fendor: maybe bad name for a flag?

ndmitchell:
Why make hslogger optional? It's not a huge dependency, and certainly making it mandatory isn't going to upset clients like ghcide.

fendor:
@ndmitchell if the additional dependencies are alright, including network and network-bsd, we can add it as a required dependency. opinion, @mpickering ?

ndmitchell:
We already have `hslogger` as a dependency due to `haskell-lsp`.

PR #70, opened 2019-10-04
Loosen lower bound of hslogger dep

fendor:
Haskell-lsp  does not define a version contraint. I dont think it is necessary for us, or is it? 
Otherwise we have to add more extra-deps to stack-*.yaml files in hie.


mpickering:
What problem is this fixing? Removing the constraint completely is not my preferred solution but i don't know what the problem is in order to suggest the correct fix.

fendor:
It would spare h-i-e to define an extra dependency for stack builds. It would also be possible to extend the version range to `>= 1.2 && < 1,4` or something similar.

fendor:
Lowered the lower bound hslogger to `>= 1.2 && < 1.4`

PR #72, opened 2019-10-13
Rename actionName of none-cradle

fendor:


mpickering:
Thanks for spotting this.

PR #74, opened 2019-10-15
Initialise plugins in frontend hook

mpickering:
The generic frontend already contains this call to do the initialisation
but because we bypassed it, the plugins were never being loaded.

There is probably more we can do to stop repeatedly loading some plugins
but until someone complains, plugins are quite rare so not going to
invest the effort.

### DIFF
--- a/src/HIE/Bios/Ghc/Load.hs
+++ b/src/HIE/Bios/Ghc/Load.hs
@@ -2,7 +2,6 @@
 {-# LANGUAGE CPP #-}
 module HIE.Bios.Ghc.Load ( loadFileWithMessage, loadFile, setTargetFiles, setTargetFilesWithMessage) where
 
-import CoreMonad (liftIO)
 import GHC
 import qualified GHC as G
 import qualified GhcMake as G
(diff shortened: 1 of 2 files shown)


isovector:
Excellent, thanks!

src/HIE/Bios/Ghc/Load.hs
```diff
@@ -114,12 +115,21 @@ collectASTs action = do
...
+  p <- G.parseModule =<< initializePluginsGhc ms
   tcm <- G.typecheckModule p
   let tcg_env = fst (tm_internals_ tcm)
   liftIO $ modifyIORef tc_ref (tcm :)
   return $ FrontendTypecheck tcg_env
 
+initializePluginsGhc :: ModSummary -> Ghc ModSummary
+initializePluginsGhc ms = do
+  hsc_env <- getSession
+  df <- liftIO $ initializePlugins hsc_env (ms_hspp_opts  ms)
```
fendor: Needs to be guarded by CPP or defined in a Compat Module

fendor:
Otherwise: the pr works as intended with Haskell-IDE-Engine!

mpickering:
I fixed CI and squashed the commits together.

PR #75, opened 2019-10-17
Cradle: don't call stack with --slient to preserve error messages

domenkozar:


mpickering:
I don't think this is going to work? Now lots of useless output will be sent to stdout which will mean it's harder to get the right arguments.

mpickering:
In order to fix this properly, stack need to provide an end point which gives tool authors the correct API.

domenkozar:
Or that --silence doesn't affect stderr. Either way, I agree this won't work.

PR #76, opened 2019-10-17
Add cabal multi cradle

mpickering:
This new cradle type allows you to easily specify all the different parts of a cabal project without 
lots of boilerplate.

```
cradle:
  cabal:
    - path: "./src"
      component: "lib:hie-bios"
    - path: "./test"
      component: "test"
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,9 @@ The complete configuration is a subset of
 cradle:
   cabal:
     component: "optional component name"
+  cabal:
+    - path: ./
+      component: "component name"
   stack:
   bazel:
   obelisk:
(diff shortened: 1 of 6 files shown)


mpickering:
I will add a test as I didn't even try the cradle after writing it.

If there are nested multi-cradles then at each nesting level the most specific path is chosen. They are not flattened together.

fendor:
>
> If there are nested multi-cradles then at each nesting level the most specific path is chosen. They are not flattened together.

Ok, that is the better behaviour imo as well, but we should document it somewhere. 
Other than that, it looks good to me, but I have yet to test the pr myself.

mpickering:
I added the test and some more documentation as you suggested.

fendor:
We need more tests. Simple tests that could be nice for multi-cradles:
Given a hie.yaml specification and a filepath to some project, what cradle would be used for the file.

PR #77, opened 2019-10-18
Only use an implicit Stack cradle if stack is executable

expipiplus1:
This could certainly be done for the other cradles too


expipiplus1:
Thanks!

PR #80, opened 2019-10-24
Correctly parse the output from the bios executable

jinwoo:
Multi-word arguments must be parsed correctly. E.g., currently `-flag
'foo bar'` is parsed as [-flag, 'foo, bar'].
It should be [-flag, 'foo bar'].

Use the shellwords package for that.

### DIFF
--- a/hie-bios.cabal
+++ b/hie-bios.cabal
@@ -53,6 +53,7 @@ Library
                         process              >= 1.6.1 && < 1.7,
                         file-embed           >= 0.0.10 && < 0.1,
                         ghc                  >= 8.2.2 && < 8.9,
+                        shellwords           >= 0.1.2 && < 0.2,
                         transformers         >= 0.5.2 && < 0.6,
                         temporary            >= 1.2 && < 1.4,
                         text                 >= 1.2.3 && < 1.3,
--- a/src/HIE/Bios/Cradle.hs
+++ b/src/HIE/Bios/Cradle.hs
@@ -22,6 +22,7 @@ import Data.FileEmbed
 import System.IO.Temp
 import Data.List
 import Data.Ord (Down(..))
+import qualified ShellWords as SW
 
 import System.PosixCompat.Files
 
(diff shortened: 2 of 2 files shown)


hie-bios.cabal
```diff
@@ -53,6 +53,7 @@ Library
                         process              >= 1.6.1 && < 1.7,
                         file-embed           >= 0.0.10 && < 0.1,
                         ghc                  >= 8.2.2 && < 8.9,
+                        shellwords           >= 0.1.2 && < 0.2,
```
mpickering: This adds a megaparsec dependency which I want to avoid.

mpickering:
The solution in the cabal case was to delimit the arguments by `\NULL`. Would that work here?

jinwoo:
> The solution in the cabal case was to delimit the arguments by `\NULL`. Would that work here?

I can do that. But wouldn't it break existing users?

Or I can write my own parsing logic here, but I guess that'll require some kind of parser combinator package anyway to make it simple.

If you're fine with breaking existing users, I can go with the `\NULL` delimiter solution.

jinwoo:
What about using a newline as the delimiter? I assume that's easier for custom bios programs to generate outputs.

```
-foo arg1 -bar 'multi word arg'
```
will become
```
-foo
arg1
-bar
multi word arg
```

hie-bios.cabal
jinwoo: Now this PR expects arguments to be newline-delimited. Each argument must be in its own line.

So the only change now is `words res` -> `lines res`.

This will break existing users.

mpickering:
Using newlines seems like quite an ingenious solution. In bash is there a convenient way to convert a list of arguments separated by spaces to one separated by newlines? I am just wondering how to convert the GHCi target in the ghc repo to work if the repl arguments target returns newline rather than space separated arguments. 

The documentation also needs updating but a lot of it is out of date anyway.

jinwoo:
Made a slight change in the README.md file.

Re: converting spaces to newlines, maybe `sed`? But if we blindly convert spaces to newlines, things will break again. Maybe convert only when the output contains no newline characters. But it can be also a problem when the output contains only one argument and it is a multi-word argument.

mpickering:
I think it will be easier to convert newlines into spaces, so I'll go that direction.

PR #85, opened 2019-11-07
Need to export CradleLoadResult

afcowie:
The return type of `getCompilerOptions` changed, so the major version of **hie-bios** needs to bump to n+1, in this case `0.3.0`

It's new return type, CradleLoadResult, needs to be exported. Done.


mpickering:
There are many changes since `0.2` which require a version bump. It's not surprising that `ghcide` won't compile against the master branch.

Issue #65, opened 2019-10-01
Add optional list of components to stack cradle options


mitchellwrosen:
This would allow you to use the `stack` cradle in a context where you may not want to load every package defined in your `stack.yaml` at once (e.g. in a big monorepo).

mpickering:
I think this is an issue with how `ghcide` uses the stack cradle. AFAIU, it just passes an empty path to the cradle which means that stack loads everything.

ndmitchell:
Currently, yes, ghcide does that. It's something that needs fixing at some point.

afcowie:
I'm not sure if this is the same issue, but I'm not sure how to get `cradle: {stack}` to load test dependencies? Running in a checkout of **ghcide** I get errors like

```
File:     ./test/exe/Main.hs
Range:    26:7-26:23
Source:   not found
Severity: DsError
Message: 
  Could not find module ‘Test.Tasty.HUnit’
  It is not a module in the current program, or in any known package.
...
Files that failed:
 * ./test/exe/Main.hs
 * ./test/src/Development/IDE/Test.hs

Completed (39 files worked, 2 files failed)

```

@ndmitchell I'm pretty sure you showed dogfooding the _ghcide_ program in your own code, so I'm missing something obvious, I assume.

jneira:
I've hit the same issue than @afcowie using directly `hie-bios` (master):

```
D:\ws\haskell\hie-bios>hie-bios check tests\ParserTests.hs
cradle
  Cradle {cradleRootDir = "D:\\ws\\haskell\\hie-bios", cradleOptsProg = CradleAction: stack}
hie-bios: <command line>: cannot satisfy -package tasty-1.2.3
    (use -v for more information)
```

But clearly it is a separate issue: `hie-bios` should be able to load all stack components before optionally only load some of them!

mpickering:
@afcowie I don't know how to reproduce your issue but I would like to if you can tell me how. 

@jneira I reproduce your issue. The problem is that `stack repl tests/ParserTests.hs` fails. If you can debug that then the stack cradle would work but it's a mystery to me why it fails.

afcowie:
@mpickering ok, let's see. As far as I can tell it's any project with e.g. a `library` _and_ a `test-suite` where the tests binaries depend on things (like say **hspec** or **hedgehog**) that the core `library` does not.

Do you want an example project you can clone/unpack and try for yourself?

mpickering:
You can now explicitly specify a component.

[...]

PR #86, opened 2019-11-07
Add --test so that test dependencies are built

afcowie:
I've been poking away at trying to solve #65 . My hypothesis is that all I need to do is add `--test --no-run-tests` (and maybe `--bench --no-run-benchmarks` for completeness) to `stackAction` and we're done?


afcowie:
(I'm really not sure at the wisdom of this, but I wanted to have a go)

mpickering:
@afcowie I don't think you are solving any problem with `hie-bios`. The issue is that `ghcide` doesn't pass a filepath when it calls the cradle. 

If you run `stack repl /path/to/test/file.hs` then it will work correctly and build test dependencies I think.

PR #87, opened 2019-11-14
API tidy up, comments, remove unused code etc

mpickering:


hie-bios.cabal
```diff
@@ -27,17 +27,16 @@ Library
                         HIE.Bios.Config
                         HIE.Bios.Cradle
                         HIE.Bios.Environment
-                        HIE.Bios.Debug
+                        HIE.Bios.Internal.Debug
```
fendor: Listed modules were sorted. If it doesnt matter, not a blocker, just sayin'

src/HIE/Bios/Ghc/Logger.hs
```diff
@@ -35,11 +31,11 @@ newtype LogRef = LogRef (IORef Builder)
...
 newLogRef = LogRef <$> newIORef id
 
-readAndClearLogRef :: Options -> LogRef -> IO String
-readAndClearLogRef opt (LogRef ref) = do
+readAndClearLogRef :: LogRef -> IO String
+readAndClearLogRef (LogRef ref) = do
     b <- readIORef ref
     writeIORef ref id
-    return $! convert opt (b [])
+    return $! unlines (b [])
```
fendor: Out of interest, I think this is one of the very few places where `($!)` is used. Is there a reason for it?

fendor:
Haddock Documentation is a bit broken. We can fix that in a follow-up PR, though.

src/HIE/Bios/Ghc/Logger.hs
mpickering: This is old code from `hpp` still so I can't comment on the reason.

PR #89, opened 2019-11-14
Fix and extend Haddock documentation

fendor:


src/HIE/Bios/Ghc/Api.hs
```diff
@@ -57,10 +57,10 @@ withGhcT body = do
 
 ----------------------------------------------------------------
 
--- | Initialise a GHC session by loading a given file into a given cradle.
+-- | Initialize a GHC session by loading a given file into a given cradle.
```
mpickering: 😱

src/HIE/Bios/Types.hs
```diff
@@ -5,45 +5,56 @@
...
 
 instance Show CradleAction where
   show CradleAction { actionName = name } = "CradleAction: " ++ name
 
-data CradleLoadResult r = CradleSuccess r -- ^ The cradle succeeded and returned these options
-                      | CradleFail CradleError -- ^ We tried to load the cradle and it failed
-                      | CradleNone -- ^ No attempt was made to load the cradle
+-- | Result of an attempt to set up a GHC session for a 'Cradle'.
+-- This is the go-to error handling mechanism. When possible, this
+-- should be preferd over throwing exceptions.
```
mpickering:
```suggestion
-- should be preferd over throwing exceptions.
++ should be preferred over throwing exceptions.
```

src/HIE/Bios/Ghc/Api.hs
fendor: Hehe, I actually wanted to ask: Are we using GB or US english here? Function names are US but most comments are GB

PR #95, opened 2019-11-19
Fix `fixImportDirs`

fendor:
Use platform independent pathSeparator instead of "/".

Add additional check to avoid expanding "-i" to "-i<pwd>"
since "-i" means: "Empty the import directory list"

Reference: https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/flags.html#finding-imports


mpickering:
thanks!

PR #97, opened 2019-12-02
Add optional component to stack cradle and multi-stack

### LINKED ISSUE #96: Changes to the stack cradle
1. Don't pass a FilePath to the cradle, rely on the magic all component logic
2. Add an option to specify a specific component, like the cabal cradle
3. Simple parsing of the ghci script to file :add and :load targets.

fendor:
Closes #96 

* [x]  Don't pass a FilePath to the cradle, rely on the magic all component logic
* [x]  Add an option to specify a specific component, like the cabal cradle
* [x] Simple parsing of the ghci script to file :add and :load targets.

### DIFF
--- a/src/HIE/Bios/Cradle.hs
+++ b/src/HIE/Bios/Cradle.hs
@@ -35,6 +35,7 @@ import qualified Data.Conduit.Combinators as C
 import qualified Data.Conduit as C
 import qualified Data.Conduit.Text as C
 import Data.Text (unpack)
+import           Data.Maybe                     ( maybeToList )
 ----------------------------------------------------------------
 
 -- | Given root\/foo\/bar.hs, return root\/hie.yaml, or wherever the yaml file was found.
@@ -73,7 +74,12 @@ getCradle (cc, wdir) = addCradleDeps cradleDeps $ case cradleType cc of
         (CradleConfig cradleDeps
           (Multi [(p, CradleConfig [] (Cabal (Just c))) | (p, c) <- ms])
         , wdir)
-    Stack -> stackCradle wdir
+    Stack mc -> stackCradle wdir mc
+    StackMulti ms ->
+      getCradle $
+        (CradleConfig cradleDeps
+          (Multi [(p, CradleConfig [] (Stack (Just c))) | (p, c) <- ms])
+        , wdir)
  --   Bazel -> rulesHaskellCradle wdir
  --   Obelisk -> obeliskCradle wdir
     Bios bios deps  -> biosCradle wdir bios deps
@@ -103,7 +109,7 @@ implicitConfig' fp = (\wdir ->
          (Bios (wdir </> ".hie-bios") Nothing, wdir)) <$> biosWorkDir fp
   --   <|> (Obelisk,) <$> obeliskWorkDir fp
   --   <|> (Bazel,) <$> rulesHaskellWorkDir fp
-     <|> (stackExecutable >> (Stack,) <$> stackWorkDir fp)
+     <|> (stackExecutable >> (Stack Nothing,) <$> stackWorkDir fp)
      <|> ((Cabal Nothing,) <$> cabalWorkDir fp)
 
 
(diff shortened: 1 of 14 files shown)


src/HIE/Bios/Cradle.hs
```diff
@@ -353,39 +362,68 @@ cabalWorkDir = findFileUpwards isCabal
...
+        return $ makeCradleResult (combineExitCodes [ex1, ex2], stde ++ stdr, ghc_args ++ pkg_ghc_args ++ targets) deps
+
+findGhciScript :: [String] -> Maybe FilePath
+findGhciScript args
+    | Just option <- find (ghciScript `isPrefixOf`) args
+    , Just flag <- stripPrefix ghciScript option
+    = Just flag
+    | otherwise = Nothing
+    where
+      ghciScript = "-ghci-script="
```
mpickering: This is a bit dodgy. I though this logic would be in the session initialization function as then you can look at the `DynFlags` field to get the ghci script file just before setting the targets.

src/HIE/Bios/Cradle.hs
```diff
@@ -353,39 +362,68 @@ cabalWorkDir = findFileUpwards isCabal
...
+
+getTargetsFromGhciScript :: FilePath -> IO [String]
+getTargetsFromGhciScript script = do
+  contents <- lines <$> readFile script
+  return
+    $ concatMap (tail {- first element is ":add" which we want to remove -}
```
mpickering: Add a comment that the lines look like `:add Foo.hs Bar.hs`.

wrappers/cabal
```diff
@@ -10,5 +10,5 @@ if [ "$1" == "--interactive" ]; then
     out "$arg"
   done
 else
-  ghc "$@"
+  ${HIE_BIOS_GHC} "$@"
```
mpickering: Does this work if there's a space in the path?

mpickering:
Great patch thanks. One structural comment about where we parse the ghci script file.

wrappers/cabal
fendor: Fair, we could also set a default variable right here, so cabal doesnt have to if not necessary.

src/HIE/Bios/Cradle.hs
fendor: I felt like here it would also be nice because we can easily debug it. It would follow the principle that the output of `hie-bios flags ...` can directly be used to compile a file.
If we are looking at the DynFlags in initSession for example, then we dont know the cradle type anymore and parse the ghci-script for every cradle. If this is desired behaviour, then it makes sense.

jneira:
This would make test components work as well (at least for stack cradle)?

fendor:
Locally, the tests succeed for me as well, I just fail to make CI happy :/

mpickering:
I was just thinking that instead of calling ghc directly for stack we should call `stack exec ghc`? That would make it also work for the docker integration I believe and avoid having to manually deal with the compiler path.

fendor:
CI is happy again! 
Turns out, there were actually mistakes in the `hie.yaml` files, they were actually using cabal and now we fix this by writing a `stack.yaml` based on the used ghc.

fendor:
I think, we do have a breaking change in hie.yaml:
```yaml
cradle:
  stack:
```
is no longer a valid yaml configuration.
My suggestion would be to lift this restriction, for cabal and stack. If there is no object, then just assume `Nothing`. @mpickering, @jneira opinion?

jneira:
Sounds good to me.
The test suite in windows continue stable: only one failing test (same than master)

tests/BiosTests.hs
```diff
@@ -6,44 +7,45 @@ import HIE.Bios
...
+        , testCaseSteps "simple-cabal-unknown-path"
+                (findCradleForModule
+                  "./tests/projects/simple-cabal/Foo.hs"
+                  (Just "./tests/projects/simple-cabal/hie.yaml")
+                )
+        ]
+      , testGroup "Loading tests" [
+        testCaseSteps "simple-cabal" $ testDirectory "./tests/projects/simple-cabal/B.hs"
+        -- The stack tests don't attempt to load the targets initially because they are not returned
+        -- by `stack repl`. They are hidden inside a GHCi script.
```
mpickering: This comment is out of date I think.

mpickering:
I think this can be merged once CI passes?

PR #103, opened 2019-12-14
Lower cabal-version and update README

fendor:


jneira:
Maybe we could mention the new env var in the README?

fendor:
Actually, I changed my mind. These env variables are subject to change, since their usage is not very consequent. I would like to overhaul them a bit after a release.
If we add them to the README, they become obserable behaviour on which one can rely on.

hie-bios.cabal
```diff
@@ -14,13 +14,54 @@ Build-Type:             Simple
...
+                        tests/projects/simple-direct/A.hs
+                        tests/projects/simple-direct/B.hs
+                        tests/projects/simple-direct/hie.yaml
+                        tests/projects/simple-stack/A.hs
+                        tests/projects/simple-stack/B.hs
+                        tests/projects/simple-stack/CHANGELOG.md
+                        tests/projects/simple-stack/Setup.hs
+                        tests/projects/simple-stack/cabal.project
+                        tests/projects/simple-stack/hie.yaml
+                        tests/projects/simple-stack/simple-stack.cabal
```
mpickering: I hate this..

Issue #102, opened 2019-12-14
Adding +RTS to ghc-options fails with an error


csasarak:
This is a duplicate of an issue in: https://github.com/digital-asset/ghcide/issues/136 

> 
> I had some `+RTS ... -RTS` options set for GHC in the `ghc-options` section of my `package.yaml` file. Running `ghcide` in that directory fails with `ghcide: target ‘+RTS’ is not a module name or a source file` The full output is:
> 
> ```
> `Step 1/6: Finding files to test in /Users/<user>/<project>
> Found 601 files
> 
> Step 2/6: Looking for hie.yaml files that control setup
> Found 1 cradle
> 
> Step 3/6, Cradle 1/1: Implicit cradle for /Users/user/project
> Cradle {cradleRootDir = "/Users/user/project", cradleOptsProg = CradleAction: stack}
> 
> Step 4/6, Cradle 1/1: Loading GHC Session
> ghcide: target ‘+RTS’ is not a module name or a source file
> ```

I am investigating and can duplicate the problem, but for some reason only with the cabal files generated by `stack`/`hpack`. I am going to keep investigating and try to cause the issue with pure cabal in addition to trying to fix it.

mpickering:
I can reproduce this by just adding `+RTS -s -RTS` to the `ghc-options` stanza in a cabal file. What difficulty are you having? 

Patch should be straightforward -- remove any RTS arguments.

csasarak:
I only see this behavior sometimes. For example, modifying `hie-bios`'s own cabal file and changing the Executable section to add `+RTS ...` bits:
`
Executable hie-bios
  Default-Language:     Haskell2010
  Main-Is:              Main.hs
  Other-Modules:        Paths_hie_bios
  ghc-options:          +RTS -s -RTS -Wall 
  HS-Source-Dirs:       exe
  Build-Depends:        base >= 4.9 && < 5
                      , directory
                      , filepath
                      , ghc
                      , hie-bios
`
And then running hie-bios (maybe this is the step I'm messing up) with:
`cabal run hie-bios -- check exe/Main.hs` 
I get no error.

fendor:
When adding to `ghc-options`, shouldnt it be `"-with-rtsopts=-s"`? Or is that equivalent?
https://www.haskell.org/cabal/users-guide/developing-packages.html?highlight=ghc%20options#pkg-field-ghc-options

ndmitchell:
[...]

PR #106, opened 2019-12-18
Strip any '+RTS ...' flags from command line options, remove GHC 8.2.2 support

csasarak:
Addresses #102 

I had a version of this in `addCmdOpts` but decided this was more appropriate. I can move it back if it isn't.

### DIFF
--- a/src/HIE/Bios/Cradle.hs
+++ b/src/HIE/Bios/Cradle.hs
@@ -288,6 +288,7 @@ processCabalWrapperArgs args =
         (dir: ghc_args) ->
             let final_args =
                     removeVerbosityOpts
+                    $ removeRTS
                     $ removeInteractive
                     $ map (fixImportDirs dir)
                     $ ghc_args
@@ -317,7 +318,7 @@ getCabalWrapperTool (ghcPath, ghcArgs) wdir = do
             let wrapper_hs = cacheDir </> wrapper_name <.> "hs"
             writeFile wrapper_hs cabalWrapperHs
             let ghc = (proc ghcPath $
-                        ghcArgs ++ ["-outputdir", tmpDir, "-o", wrapper_fp, wrapper_hs])
+                        ghcArgs ++ ["-rtsopts=ignore", "-outputdir", tmpDir, "-o", wrapper_fp, wrapper_hs])
                         { cwd = Just wdir }
             readCreateProcess ghc "" >>= putStr
         return wrapper_fp
@@ -345,6 +346,15 @@ cabalAction work_dir mc l _fp = do
 
 removeInteractive :: [String] -> [String]
 removeInteractive = filter (/= "--interactive")
+
+-- Strip out any ["+RTS", ..., "-RTS"] sequences in the command string list.
+removeRTS :: [String] -> [String]
+removeRTS ("+RTS" : xs)  =
+  case dropWhile (/= "-RTS") xs of
+    [] -> []
+    (_ : ys) -> removeRTS ys
+removeRTS (y:ys)         = y : removeRTS ys
+removeRTS []             = []
 
 removeVerbosityOpts :: [String] -> [String]
 removeVerbosityOpts = filter ((&&) <$> (/= "-v0") <*> (/= "-w"))
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,8 +37,6 @@ matrix:
       addons: {apt: {packages: [ghc-ppa-tools,cabal-install-2.4,ghc-8.6.5], sources: [hvr-ghc]}}
     - compiler: "ghc-8.4.4"
       addons: {apt: {packages: [ghc-ppa-tools,cabal-install-2.4,ghc-8.4.4], sources: [hvr-ghc]}}
-    - compiler: "ghc-8.2.2"
-      addons: {apt: {packages: [ghc-ppa-tools,cabal-install-2.4,ghc-8.2.2], sources: [hvr-ghc]}}
 
 before_install:
   - HC=/opt/ghc/bin/${CC}
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,7 +4,6 @@ environment:
  # - GHCVER: 8.8.1 Disabled until unix-compat supports ghc 8.8.1, see https://github.com/jacobstanley/unix-compat/pull/42 for example
   - GHCVER: 8.6.5
   - GHCVER: 8.4.4
-  - GHCVER: 8.2.2
 install:
   - git submodule update --init --recursive
   - choco install -y haskell-stack
--- a/hie-bios.cabal
+++ b/hie-bios.cabal
@@ -66,7 +66,7 @@ Extra-Source-Files:     ChangeLog
                         tests/projects/simple-stack/simple-stack.cabal
 
 
-tested-with: GHC==8.6.5, GHC==8.4.4, GHC==8.2.2
+tested-with: GHC==8.6.5, GHC==8.4.4
 
 Library
   Default-Language:     Haskell2010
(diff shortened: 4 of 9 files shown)


mpickering:
Can you add a test please?

csasarak:
Yes

csasarak:
With respect to testing, I make new projects for each of stack/cabal and test against those, or add but I could also add a trivial `+RTS -A1m -RTS` to the existing ones. Is that sufficient? 

I can also add some more extensive tests of the function itself with different inputs

mpickering:
Adding the options to the existing tests would be fine, adding a new test would also be fine, it doesn't matter too much which you choose.

src/HIE/Bios/Cradle.hs
```diff
@@ -345,6 +346,16 @@ cabalAction work_dir mc l _fp = do
...
+-- Strip out any ["+RTS", ..., "-RTS"] sequences in the command string list.
+removeRTS :: [String] -> [String]
+removeRTS  = snd . foldr processPath (False, [])
+  where processPath p (inRTS, paths) =
+          case (inRTS, p) of
+            (_, "+RTS") -> (False, paths)
+            (_, "-RTS") -> (True, paths)
+            (True, _)   -> (inRTS, paths)
+            (_, p')     -> (False, p' : paths)
+
```
jinwoo: Sorry for a drive-by comment. Would code like this be simpler?

```haskell
removeRTS :: [String] -> [String]
removeRTS [] = []
removeRTS ("+RTS" : xs) =
  case dropWhile (/= "-RTS") xs of
    [] -> []
    (_ : ys) -> removeRTS ys
```

csasarak:
I'm not at my computer at the moment, but as written i don't think this would work. The argument sequences also have other arguments in them besides the '+RTS...' bits which aren't necessarily at the beginning or end of the list.

If i'm understanding the proposal, I *think* your code will work with the addition of this case after your code.

```
removeRTS (y:ys) = y : removeRTS ys
```

I like the readability of what you wrote a lot and can change/test it again later. Thank you for having a look!


On December 19, 2019 4:58:37 PM EST, Jinwoo Lee <notifications@github.com> wrote:
>jinwoo commented on this pull request.
>
>
>
>> +-- Strip out any ["+RTS", ..., "-RTS"] sequences in the command
>string list.
>+removeRTS :: [String] -> [String]
>+removeRTS  = snd . foldr processPath (False, [])
>+  where processPath p (inRTS, paths) =
>+          case (inRTS, p) of
>+            (_, "+RTS") -> (False, paths)
>+            (_, "-RTS") -> (True, paths)
>+            (True, _)   -> (inRTS, paths)
>+            (_, p')     -> (False, p' : paths)
>+
>
>Sorry for a drive-by comment. Would code like this be simpler?
>
>```haskell
>removeRTS :: [String] -> [String]
>removeRTS [] = []
>removeRTS ("+RTS" : xs) =
>  case dropWhile (/= "-RTS") xs of
>    [] -> []
>    (_ : ys) -> removeRTS ys
>```

-- 
Chris

src/HIE/Bios/Cradle.hs
csasarak: If i'm understanding the proposal, I *think* your code will work with the addition of this case after your code.

```
removeRTS (y:ys) = y : removeRTS ys
```

I like the readability of what you wrote a lot and can change/test it again later. Thank you for having a look!

jinwoo: Oops. Yeah, I forgot to add that case :)

csasarak: A new version is up.

jinwoo:
Another drive-by comment :)

Per https://downloads.haskell.org/ghc/latest/docs/html/users_guide/runtime_control.html#rts-opts-cmdline, "No -RTS option is required if the runtime-system options extend to the end of the command line". It'd be good to add a test case like that too.

csasarak:
@jinwoo I looked at this situation and cabal doesn't seem to build anything without the `-RTS`. When I checked, it's because additional non-rts ghc options were being added after `+RTS`. In my test it was adding `-hide-all-packages` which according to the [manual](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/packages.html#ghc-flag--hide-all-packages) is always passed by `cabal`. 

Even if GHC can accept the `+RTS ...` form on the command line, I don't think it makes sense in the context of a cabal file. I don't know of any way to control the order of options that cabal uses when invoking GHC, which would be necessary for the `+RTS ...` form to work reliably.

I couldn't think of what the test case for that would achieve then, since the cabal file seems to be unusable if you leave off the `-RTS` option. A cabal file where `hie-bios` works but cabal doesn't is maybe not an important case to support. But I want to hear if you agree with that, maybe there's a good reason for the additional test that I'm not thinking of. Or I could ask the cabal devs if that situation is valid or there's a way to mitigate it.

jinwoo:
Oh I see. I had to do something similar in our environment and because we don't use cabal, I didn't realize it doesn't happen with cabal. Sorry about that.

I also realize that this change affects only cabal cradles. Why not make this change for all types of cradles? I guess it would benefit all of them.

Having said that, as I'm not the owner of this project, I'll just shut up now :)

Thank you for making this change!

csasarak:
@jinwoo I see what you're saying and realize that I ignored the `bios` cradle type when writing tests. To make sure I've gotten everything:

* Stack: If you use stack then there's still a cabal file, and as far as I can tell it a stack cradle will still run things through `cabalWrapperArgs`. So stack cradles are supported by this code. Like regular cabal projects, stack won't build a project with `+RTS ...` form `ghc-options` in the cabal file.  There is a test as part of this PR for stack too.

* Cabal: This is supported, and exhibits the behavior where a project won't build if `ghc-options` has a `+RTS ...` form argument. Since cabal doesn't support it, I didn't worry about this case. 

* BIOS: This one I didn't have a test for, so thank you for prompting me to check. I don't see that anything was broken if you add +RTS options to the the script for this one. 

Can you let me know if there's a case here that isn't supported?

csasarak:
So the current status of this is that the appveyor tests failed because the test RTS opts were being passed into the wrapper, which was exiting with an error because the RTS options GHC defaults to are 'some'. I changed the wrapper compilation code to include `-rtsopts=ignore` which just passes RTS options unchanged into the wrapper. It works for GHCs > 8.2.2, 8.2.2 and below don't seem to have an 'ignore' option. 

Setting `rtsopts=all` is wrong because it means the wrapper is eating the RTS options and not passing them to GHC underneath.

mpickering:
We can probably just remove 8.2.2 support if it's causing issues. I think h-i-e or ghcide support that version.

csasarak:
I think the simplest fix would be dropping support for 8.2.2 and using `-rtsopts=ignore`. I could turn on `rtsopts=all` for the `cabal.hs` program, but that would mean on windows the wrapper program catches those RTS opts and not the underlying call to `HIE_BIOS_GHC` which is what happens on linux. 

Can you clarify if this is correct: cabal or the `cabal.hs` program is receiving the same ghc-options as whatever component it's building. I'm guessing that that's because maybe a user might want some warnings for some components and not others, and the checks can be made to match what cabal would do at build-time for a given component. This is where the appveyor tests are failing I think, because the windows wrapper program doesn't accept the full complement of RTS options while cabal and stack do. 

Do you think that adding `-rtsopts=ignore` sufficient, or is that a bandaid?

mpickering:
I think dropping support for 8.2.2 and using `-rtsopts=ignore` is a good solution. 

I don't understand your question sorry, `cabal` uses the compiled `cabal.hs` executable instead of `ghc` when invoked by `hie-bios`, this allows us to intercept the options it supplies to GHC to start a session.

csasarak:
> I think dropping support for 8.2.2 and using `-rtsopts=ignore` is a good solution.

Sounds good, I will do that

> I don't understand your question sorry, `cabal` uses the compiled `cabal.hs` executable instead of `ghc` when invoked by `hie-bios`, this allows us to intercept the options it supplies to GHC to start a session.

I think I was just chasing my tail. I have a better understanding of how things work thanks to your comment.

csasarak:
I think this is finished, but appveyor had a server error. I will try triggering a build again later.

mpickering:
Needs a rebase and squash.

csasarak:
@mpickering Let me know if there's anything I can do to further assist here. Thanks for your work on this project!

PR #112, opened 2019-12-18
Fix wrapper to default to no ghc arguments and execute compilation in cradle dir

fendor:
In the wrapper, if the env variable `HIE_BIOS_GHC_ARGS` is set to `""`, in windows, this seems to throw an exception. Fix it by using `lookupEnv` and adding a default value.

Also, mainly for stack exec, execute `stack exec ghc --` in the working directory of the cradle, to avoid that the global project is used for compilation.

Manually tested on windows for cabal in wsl, stack not yet.
EDIT: apparently I wasnt using the wsl but rather in mingw32? or cygwin? Not sure, to many artifacts of previous attempts to turn the windows device into a dev machine.

### DIFF
--- a/src/HIE/Bios/Cradle.hs
+++ b/src/HIE/Bios/Cradle.hs
@@ -302,8 +302,8 @@ type GhcProc = (FilePath, [String])
 -- generate a fake GHC that can be passed to cabal
 -- when run with --interactive, it will print out its
 -- command-line arguments and exit
-getCabalWrapperTool :: GhcProc -> IO FilePath
-getCabalWrapperTool (ghcPath, ghcArgs) = do
+getCabalWrapperTool :: GhcProc -> FilePath -> IO FilePath
+getCabalWrapperTool (ghcPath, ghcArgs) wdir = do
   wrapper_fp <-
     if isWindows
       then do
@@ -317,7 +317,7 @@ getCabalWrapperTool (ghcPath, ghcArgs) = do
             writeFile wrapper_hs cabalWrapperHs
             createDirectoryIfMissing True cacheDir
             let ghc = (proc ghcPath $ ghcArgs ++ ["-o", wrapper_fp, wrapper_hs])
-                        { cwd = Just (takeDirectory wrapper_hs) }
+                        { cwd = Just wdir }
             readCreateProcess ghc "" >>= putStr
         return wrapper_fp
       else writeSystemTempFile "bios-wrapper" cabalWrapper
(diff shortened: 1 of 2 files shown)


fendor:
Now also manually tested with stack and seemed to work.

jneira:
I've tested it locally and my error related with `Cannot decode byte '\xbd'` is gone and `hie-bios check`works correctly

src/HIE/Bios/Cradle.hs
```diff
@@ -317,7 +317,7 @@ getCabalWrapperTool (ghcPath, ghcArgs) = do
             writeFile wrapper_hs cabalWrapperHs
             createDirectoryIfMissing True cacheDir
             let ghc = (proc ghcPath $ ghcArgs ++ ["-o", wrapper_fp, wrapper_hs])
-                        { cwd = Just (takeDirectory wrapper_hs) }
+                        { cwd = Just wdir }
```
jneira: Out of curiosity: why te compilation is in the working dir if  the wrapper are gonna be as before?

fendor: Because we are using stack now to compile wrapper, since some users might not have a raw ghc installed. But if you execute stack exec ghc in the appdata something directory, we are not picking up the current stack directory, but the global one. Thus, execute stack ghc in the cradle directory, to pick up the correct ghc version

fendor:
I can not explain the error relating the byte, so I don't know if this actually fixes it

src/HIE/Bios/Cradle.hs
mpickering: Does this produce `.hi` and `.o` files in the working directory? I would guess it's ok still.

fendor: Good point, no idea, will test it.

fendor: Compilation artifacts are in the same folder as the location of the input file. In this case, the compilation artifacts are found in %TMP%.

mpickering:
Looks right to me but we need to add a step to CI to test the wrapper still.

ndmitchell:
I had an error with a weird byte encoding as a result of picking up a really old GHC with a recent Stack. Such combinations are poorly tested - I suspect this change will have helped fix it.

fendor:
Should we just merge this and we add a test for windows later? Or should we add a test for it immediately?

ndmitchell:
Has someone actually tested on Windows? I'd suggest merging ASAP if it's had a manual test, then making a quick release (it's pretty bust for me right now), then looking into a CI test (I use Appveyor).

mpickering:
I believe @fendor tested it on windows.

fendor:
yeah, i tested it on windows 10 manually for both a cabal and stack cradle. `hie-bios check src/HIE/Bios.hs` worked for both of them, e.g. nothing by default but showed the error if I introduced one.

mpickering:
I did a new release.

ndmitchell:
Thanks - I confirm the new release all works with my testing.

PR #117, opened 2019-12-20
Make wrapper name unique using its source hash

### LINKED ISSUE #114: Make wrapper name unique
* To avoid false negatives when testing different commits with the same numeric versions
* The unique suffix could be the hash of the wrapper source file

jneira:
* Before this the wrapper used the project version to cache the wrapper executable
* That led to false negatives testing the compilation and execution of the wrapper using the `bios-test` test suite cause the tests reused a previous wrapper executable between commits of the same version
  * Devs had to remember delete manually `%XDG_DIRECTORY%\hie-bios` to avoid it
* I've used `GHC.Fingerprint` to hash the source file and cache the executable using  its source (the most accurate one although non semantic changes will cause a cache miss)
  * And i've put the source near the executable to be able to check it quicky without have to go to hie-bios source 

* I've cherry-picked my changes onto `0.3.0` and `0.3.1` to check the test suite with them:
  * `0.3.0`: https://ci.appveyor.com/project/jneira/hie-bios/builds/29673096/job/9x5k66k9ittfyvy0
  * `0.3.1`: https://ci.appveyor.com/project/jneira/hie-bios/builds/29673244/job/nxkbys8o1gt737vq
* In both cases the test suite fails. 
  * To be fair it had failed in the windows ci without these changes cause we are not caching `%XDG_DIRECTORY%\hie-bios` between ci builds and the wrapper would be compiled every time.
  * But if we start caching `%XDG_DIRECTORY%\hie-bios` in ci for some reason it will suffer the false negatives as well
* I am using a new en var `%HIE_BIOS_CACHE_DIR%` to let users change the default one (without change the general `%XDG_CACHEDIR%`). It is not strictly necessary but it is a common practice to avoid issues with file permissions (in some envs the creation of executable files inside user directory is forbidden f.e.) 
  
Fixes #114

### DIFF
--- a/src/HIE/Bios/Environment.hs
+++ b/src/HIE/Bios/Environment.hs
@@ -1,5 +1,5 @@
 {-# LANGUAGE RecordWildCards, CPP #-}
-module HIE.Bios.Environment (initSession, getSystemLibDir, addCmdOpts) where
+module HIE.Bios.Environment (initSession, getSystemLibDir, getCacheDir, addCmdOpts) where
 
 import CoreMonad (liftIO)
 import GHC (DynFlags(..), GhcLink(..), HscTarget(..), GhcMonad)
@@ -13,6 +13,7 @@ import Control.Monad (void)
 import System.Process (readProcess)
 import System.Directory
 import System.FilePath
+import System.Environment (lookupEnv)
 
 import qualified Crypto.Hash.SHA1 as H
 import qualified Data.ByteString.Char8 as B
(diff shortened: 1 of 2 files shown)


src/HIE/Bios/Environment.hs
```diff
@@ -76,7 +77,11 @@ however, it's not really necessary as
 >   when res (removeDirectoryRecursive cd)
 -}
 getCacheDir :: FilePath -> IO FilePath
-getCacheDir fp = getXdgDirectory XdgCache (cacheDir </> fp)
+getCacheDir fp = do
+  mbEnvCacheDirectory <- lookupEnv "HIE_BIOS_CACHE_DIR"
+  cacheBaseDir <- maybe (getXdgDirectory XdgCache cacheDir) return
+                         mbEnvCacheDirectory
+  return (cacheBaseDir </> fp)
```
fendor: Documentation for this function please! 
Also, `getCacheDir` is probably a misnomer.

jneira: Mmm it retrieves the cache directory used by the lib, `getHieBiosCacheDir` would be better or you are thinking in another deficiency?

fendor: Or maybe a misnomer... Mainly adding documentation for the env variable and so on suffices for me.

PR #119, opened 2019-12-27
Add explicit cradle predicates and multi cradle depend on its cradles

fendor:
This changes multi-cradles to have the name `stack` or `cabal` if their hie.yaml specification solely consists of `stack` or `cabal` and `none` cradles. To make stuff easier, we add `is*Cradle` predicates to the API. Users of this library can make use of them to obtain the cradle-type after it has been parsed.

Fixes https://github.com/haskell/haskell-ide-engine/issues/1490

Also, changes the `ActionName` to be an extensible ADT.

### DIFF
--- a/src/HIE/Bios/Cradle.hs
+++ b/src/HIE/Bios/Cradle.hs
@@ -13,12 +13,14 @@ module HIE.Bios.Cradle (
     , isNoneCradle
     , isMultiCradle
     , isDefaultCradle
+    , isOtherCradle
   ) where
 
 import Control.Exception (handleJust)
 import System.Process
 import System.Exit
-import HIE.Bios.Types
+import HIE.Bios.Types hiding (ActionName(..))
+import qualified HIE.Bios.Types as Types
 import HIE.Bios.Config
 import HIE.Bios.Environment (getCacheDir)
 import System.Directory hiding (findFile)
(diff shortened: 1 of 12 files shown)


src/HIE/Bios/Cradle.hs
```diff
@@ -260,13 +291,57 @@ biosAction wdir bios bios_deps l fp = do
 -- Cabal Cradle
 -- Works for new-build by invoking `v2-repl` does not support components
 -- yet.
+cabalCradleName :: String
```
jneira: It seems to me that those function ask to be `cradleName :: CradleType -> String` and `isCradleType :: Cradle -> CradleType -> Bool` (so we can check with ```myCradle `isCradleType` StackCradle```)
Or even add `name` as a field to `CradleType` type (although it would suppose more radical changes)

fendor: I dont like the idea of exposing the Configuration type, if we want to edit that, we are always having breaking changes.
Moreover, it is a lousy pattern match ADT, since we have `StackMulti [(FilePath, CradleConfig)]`. 
However, I like the idea to have a ADT like:
```haskell
data CradleType 
  = Cabal
  | Stack
  | Direct
  ...
  | Other String
```
Which prepares `hie-bios` to be more extensible.

fendor:
cc @cocreature, @ndmitchell 
This patch would change the type of `Cradle` to `Cradle a` and `actionName :: String` changes to `ActionName a`. For you the main difference would be that you would have to change occurrences of either `Cradle` or `ActionName` to `Cradle Void` and `ActionName Void` respectively, afaict. Would you be fine with this PR, or do you have counter arguments, or other suggestions?
Based on this PR, I plan a follow up PR for hooking into the parsing of `hie.yaml`, e.g. extending `Config` to allow specifying a custom cradle.

ndmitchell:
Doesn't seem too big a deal for the changes in ghcide.

PR #123, opened 2020-01-05
Pass filepath to cabal v2-repl when getting flags

bubba:
Instead of component. This lets Cabal figure out what component the module is in and should give more reliable flags for modules living outside of the specified component.

### DIFF
--- a/src/HIE/Bios/Cradle.hs
+++ b/src/HIE/Bios/Cradle.hs
@@ -328,10 +328,9 @@ getCabalWrapperTool (ghcPath, ghcArgs) wdir = do
   return wrapper_fp
 
 cabalAction :: FilePath -> Maybe String -> LoggingFunction -> FilePath -> IO (CradleLoadResult ComponentOptions)
-cabalAction work_dir mc l _fp = do
+cabalAction work_dir _mc l fp = do
   wrapper_fp <- getCabalWrapperTool ("ghc", []) work_dir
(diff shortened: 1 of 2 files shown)


bubba:
Related to https://github.com/haskell/haskell-ide-engine/pull/1526

src/HIE/Bios/Cradle.hs
```diff
@@ -328,10 +328,9 @@ getCabalWrapperTool (ghcPath, ghcArgs) wdir = do
   return wrapper_fp
 
 cabalAction :: FilePath -> Maybe String -> LoggingFunction -> FilePath -> IO (CradleLoadResult ComponentOptions)
-cabalAction work_dir mc l _fp = do
+cabalAction work_dir _mc l fp = do
   wrapper_fp <- getCabalWrapperTool ("ghc", []) work_dir
-  let cab_args = ["v2-repl", "--with-compiler", wrapper_fp]
-                  ++ [component_name | Just component_name <- [mc]]
+  let cab_args = ["v2-repl", "--with-compiler", wrapper_fp, fixTargetPath fp]
```
fendor: How about the following behaviour:
If `mc` is `Nothing`, then we pass the FilePath, but otherwise, we use the given component? The same could be applied for the stack cradle.

mpickering:
This seems like a good idea but I haven't tested the patch at all and won't have time to in the foreseeable future.

PR #127, opened 2020-01-13
Catch permission errors in cradle discovery

### LINKED ISSUE #126: hie.yaml discovery logic does not account for parent directories with no read permissions
I originally reported this issue as [ghcide#314](https://github.com/digital-asset/ghcide/issues/314) but was advised there that the error probably originates in logic implemented by `hie-bios` instead. 

The issue is that we are using `ghcide` on a shared system where users don't have read permissions for a parent directory of their home directories. This results in the following crash during step 3 of the startup of `ghcide`:

```
Step 3/6, Cradle 1/1: Implicit cradle for /[directory without read permissions]/[username]/[project]
ghcide: /[directory without read permissions]: getDirectoryContents:openDirStream: permission denied (Permission denied)
```

This happened with `ghcide` compiled using the current HEAD at the time (a0aa013) which appears to be using `hie-bios-0.3.2`.

A workaround is to have an explicit hie.yaml file in the current directory.

fendor:
Closes #126

### DIFF
--- a/src/HIE/Bios/Cradle.hs
+++ b/src/HIE/Bios/Cradle.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TupleSections #-}
 {-# LANGUAGE BangPatterns #-}
 module HIE.Bios.Cradle (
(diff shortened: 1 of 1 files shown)


mpickering:
Do we catch, or check for permissions before attempting the read?

fendor:
No, in this implementation, we try to read the directory and catch the exception if there is a permission error. Then we continue to look in the parent directory.

PR #129, opened 2020-01-22
Expose initSessionWithMessage

Avi-D-coder:
Needed by https://github.com/haskell/haskell-ide-engine/pull/1589

Github and vim automatically added the newline so I gave up.

### DIFF
--- a/src/HIE/Bios/Ghc/Api.hs
+++ b/src/HIE/Bios/Ghc/Api.hs
@@ -3,6 +3,7 @@
 module HIE.Bios.Ghc.Api (
     initializeFlagsWithCradle
   , initializeFlagsWithCradleWithMessage
+  , initSessionWithMessage
   , G.SuccessFlag(..)
   -- * Utility functions for running the GHC monad and implementing internal utilities
   , withGHC


fendor:
Thanks!

fendor:
Actually, now I am not entirely sold on the idea to just exposing `initSessionWithMessage`. Maybe we could change the type of `initializeFlagsWithCradleWithMessage` to also return the `ComponentOptions` if there are any? Otherwise the function `initializeFlagsWithCradleWithMessage` is rather useless, esepcially for Haskell IDE Engine, and I think we are the only users of that function.

Would you mind brainstorming a better API for `initializeFlagsWithCradleWithMessage`?

What do you think about:
```haskell
initializeFlagsWithCradleWithMessage ::
  GhcMonad m
  => Maybe G.Messager
  -> FilePath -- ^ The file we are loading the 'Cradle' because of
  -> Cradle   -- ^ The cradle we want to load
  -> m (CradleLoadResult (ComponentOptions, m G.SuccessFlag)) -- ^ Whether we actually loaded the cradle or not.
```
Disadvantage: looks dirty
Advantage: Does not force users of the lib to re-implement initializeFlagsWithCradleWithMessage  with the newly exposed `initSessionWithMessage`.

Avi-D-coder:
@fendor If we are truly the only user of this function we can change the return type and willing to make a breaking change, this would also get ride of the unreachable case (https://github.com/haskell/haskell-ide-engine/pull/1589#discussion_r369870452). I was working under the erroneous assumption that a breaking change was not an option.

Ghcide does not use `initializeFlagsWithCradleWithMessage`, are there any other known users of hie-bios?

fendor:
Ghcide uses nothing of the HIE.Bios.Ghc namespace and I am not aware of other big users. Moreover, next release will have breaking changes regarding the Cradle type anyways

ndmitchell:
Breaking changes are not a big deal to ghcide - we expect them at this stage in the maturity cycle.

PR #131, opened 2020-01-23
Implement hook into config parser

fendor:
Allow an hook into the configs parser to make hie-bios more extensible.


fendor:
cc @ndmitchell @cocreature 
This PR allows to extend a `hie.yaml` by custom cradles.
Example:
```yaml
cradle:
  other:
    cabal-helper:
      stack:
```

Theoretically, this should not be an observable change for you.

cocreature:
looks reasonable to me :+1:

PR #136, opened 2020-01-26
Add a mechanism to prevent hie-bios from searching upwards

bubba:
Changes the logic of `findFileUpwards` to stop at any directories that contain the file `.hie-bios.stop`. This should be useful for testing different cradle setups.


src/HIE/Bios/Cradle.hs
```diff
@@ -568,8 +568,7 @@ rulesHaskellAction work_dir fp = do
 obeliskWorkDir :: FilePath -> MaybeT IO FilePath
 obeliskWorkDir fp = do
   -- Find a possible root which will contain the cabal.project
-  wdir <- findFileUpwards (== "cabal.project") fp
-  -- Check for the ".obelisk" folder in this directory
+  wdir <- findFileUpwards (== "cabal.project") fp -- Check for the ".obelisk" folder in this directory
```
bubba:
```suggestion
  wdir <- findFileUpwards (== "cabal.project") fp
  -- Check for the ".obelisk" folder in this directory
```

bubba:
@jneira @fendor Turns out I need this again for testing in hls 😅

alanz:
I previously worked around this by physically copying the directory to be tested to a `withTempDir` location.

But being able to limit scope is a good thing, e.g. trying to work on checked-out ghcide submodule of hls, only.

fendor:
Out of interest, what happens if the `.hie-bios.stop` file is in the same dir as `hie.yaml`?

bubba:
@fendor It uses the `hie.yaml` in that same directory, but if there wasn't one there then it wouldn't search any further

PR #141, opened 2020-01-27
WIP: Change `loadCradle` to create `Cradle ()`

fendor:
Dont use Void as it sometimes is confusing.


mpickering:
New version released with the correct `Void` interface.

PR #161, opened 2020-03-08
Relax the extra bounds now 1.7 is out with no breaking changes for us

ndmitchell:


fendor:
Would you need a revision on hackage as well? I dont think we have any new features that would require a release.

ndmitchell:
If we could have a Hackage revision that would be great!

fendor:
Done https://hackage.haskell.org/package/hie-bios-0.4.0

ndmitchell:
Thanks!

PR #163, opened 2020-03-17
Fix parsing of spaces in targets ghci-scripts

### LINKED ISSUE #162: Ghci Script Target parses space in Filepath with space
Issue in hie: https://github.com/haskell/haskell-ide-engine/issues/1669

Example: some ghci script:
```
:add Lib Lib2 "/home/user/Has kell/Test.hs"
``` 
is wrongfully parsed into targets: `["Lib", "Lib2", "\"home/user/Has", "kell/Test.hs\""]`.

Offending code: https://github.com/mpickering/hie-bios/blob/9fd8af352317490f1c68cb628b2ce2a8733b63e5/src/HIE/Bios/Environment.hs#L223

fendor:
fixes #162

### DIFF
--- a/src/HIE/Bios/Environment.hs
+++ b/src/HIE/Bios/Environment.hs
@@ -228,8 +228,8 @@ getTargetsFromGhciScript script = do
   let parseGhciLine = concatMap fst . filter (null . snd) . readP_to_S parser
   return $ concatMap parseGhciLine contents
 
--- |This parser aims to parse targets that are separated by space
--- and filepaths that are encloded by double-quotes.
+-- |This parser aims to parse targets and double-quoted filepaths that are separated by spaces
+-- and prefixed with the literal ":add"
 --
 -- >>> filter (null . snd) $ readP_to_S parser ":add Lib Lib2"
 -- [(["Lib","Lib2"],"")]
--- a/src/HIE/Bios/Environment.hs
+++ b/src/HIE/Bios/Environment.hs
@@ -241,7 +241,7 @@ getTargetsFromGhciScript script = do
 -- [(["Lib","Lib2","Test Exa\"mple.hs"],"")]
 parser :: ReadP [String]
 parser = do
-  _ <- string ":add "
+  _ <- string ":add" <* space1
   scriptword `sepBy` space1
 
 space1 :: ReadP [Char]
--- a/src/HIE/Bios/Environment.hs
+++ b/src/HIE/Bios/Environment.hs
@@ -19,7 +19,9 @@ import qualified Crypto.Hash.SHA1 as H
 import qualified Data.ByteString.Char8 as B
 import Data.ByteString.Base16
 import Data.List
-
+import Data.Char (isSpace)
+import Control.Applicative ((<|>))
+import Text.ParserCombinators.ReadP
 import HIE.Bios.Types
 import HIE.Bios.Ghc.Gap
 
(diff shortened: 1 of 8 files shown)


src/HIE/Bios/Environment.hs
```diff
@@ -223,7 +225,42 @@ disableOptimisation df = updOptLevel 0 df
...
   contents <- lines <$> readFile script
-  return
-    $ concatMap (tail {- first element is ":add" which we want to remove -}
-                      . words)
-    $ filter (":add" `isPrefixOf`) contents
+  let parseGhciLine = concatMap fst . filter (null . snd) . readP_to_S parser
+  return $ concatMap parseGhciLine contents
+
+-- |This parser aims to parse targets that are separated by space
+-- and filepaths that are encloded by double-quotes.
```
fendor:
```suggestion
-- |This parser aims to parse targets and double-quoted filepaths that are separated by spaces
-- and prefixed with the literal ":add"
```

src/HIE/Bios/Environment.hs
```diff
@@ -223,7 +225,42 @@ disableOptimisation df = updOptLevel 0 df
 getTargetsFromGhciScript :: FilePath -> IO [String]
 getTargetsFromGhciScript script = do
   contents <- lines <$> readFile script
-  return
-    $ concatMap (tail {- first element is ":add" which we want to remove -}
-                      . words)
-    $ filter (":add" `isPrefixOf`) contents
+  let parseGhciLine = concatMap fst . filter (null . snd) . readP_to_S parser
```
mpickering: Should it perhaps be an error if the parse fails?

src/HIE/Bios/Environment.hs
```diff
@@ -223,7 +225,42 @@ disableOptimisation df = updOptLevel 0 df
...
+-- [(["Lib","Lib2"],"")]
+--
+-- >>> filter (null . snd) $ readP_to_S parser ":add Lib Lib2 \"Test Example.hs\""
+-- [(["Lib","Lib2","Test Example.hs"],"")]
+--
+-- >>> filter (null . snd) $ readP_to_S parser ":add Lib Lib2 \"Test Exa\\\"mple.hs\""
+-- [(["Lib","Lib2","Test Exa\"mple.hs"],"")]
+parser :: ReadP [String]
+parser = do
+  _ <- string ":add "
```
mpickering: `string ":add" <* space1`

src/HIE/Bios/Environment.hs
```diff
@@ -223,7 +225,42 @@ disableOptimisation df = updOptLevel 0 df
...
+    manyTill (escaped '"' <|> anyToken) $ char '"'
+
+escaped :: Char -> ReadP Char
+escaped c = c <$ string ("\\" <> [c])
+
+value :: ReadP String
+value = many1 (satisfy (not . isSpace))
+
+anyToken :: ReadP Char
+anyToken = satisfy $ const True
```
mpickering: Pretty sure that not all tokens are valid here.

mpickering:
Thanks, left comments but an improvement anyway.

src/HIE/Bios/Environment.hs
fendor: I dont think so, since an example ghci-script from stack looks like:
```
:add A B
:module + B
```
We only want to successfully parse the first line. 
However, we could fail only for lines beginning with `:add`.

src/HIE/Bios/Environment.hs
fendor: True, only those that are valid in module names. But I dont think we want to validate these module names here and when attempting to load, we crash anyways.
But we can change it.

PR #175, opened 2020-05-03
Avoid error in windows due to temp file being locked

jneira:
* In the process to use hie-bios on ghc i hit a problem when the hie-bios script tried to write in `%HIE_BIOS_OUTPUT$`: the file was locked by the executable and the script could not write in it
  * As usual in posis systems the lock works in a way the process started by hie-bios can write 
* Now the file es created but it is not opened until the process has used it. It is not deleted auto when it finishes but i thin kit is no a big problem, being a temporal file (but i could add a deleting)
* I've changed the handling of %HIE_BIOS_OUTPUT$` to honour the possible previous value set by the user: maybe it can be used if users pnly have permissions over concrete directories or they want to keep the compiler args in a file controled by them

### DIFF
--- a/src/HIE/Bios/Cradle.hs
+++ b/src/HIE/Bios/Cradle.hs
@@ -660,13 +660,10 @@ readProcessWithOutputFile l ghcProc work_dir fp args = do
     -- Windows line endings are not converted so you have to filter out `'r` characters
     let  loggingConduit = (C.decodeUtf8  C..| C.lines C..| C.filterE (/= '\r')  C..| C.map T.unpack C..| C.iterM l C..| C.sinkList)
     (ex, stdo, stde) <- sourceProcessWithStreams process mempty loggingConduit loggingConduit
-    existsFile <- doesFileExist output_file
-    res <- if existsFile
-             then withFile output_file ReadMode $ \handle -> do
-                    hSetBuffering handle LineBuffering
-                    !res <- force <$> hGetContents handle
-                    return res
-             else return ""
+    res <- withFile output_file ReadMode $ \handle -> do
+             hSetBuffering handle LineBuffering
+             !res <- force <$> hGetContents handle
+             return res
 
     return (ex, stdo, stde, lines (filter (/= '\r') res))
 
@@ -676,11 +673,8 @@ readProcessWithOutputFile l ghcProc work_dir fp args = do
         let mbHieBiosOut = lookup hieBiosOutput env
         case mbHieBiosOut of
           Just file@(_:_) -> action file
-          _ -> withTempHieBiosOutput action
-
-      withTempHieBiosOutput action = 
-        withSystemTempDirectory "hie-bios" $ 
-          \ tmpDir -> action $ tmpDir </> "output"
+          _ -> withSystemTempFile "hie-bios" $ 
+                 \ file h -> hClose h >> action file
 
       hieBiosGhc = "HIE_BIOS_GHC"
       hieBiosGhcArgs = "HIE_BIOS_GHC_ARGS"
--- a/src/HIE/Bios/Cradle.hs
+++ b/src/HIE/Bios/Cradle.hs
@@ -645,7 +645,8 @@ readProcessWithOutputFile l ghcProc work_dir fp args = do
             ( fromMaybe "ghc" (lookup hieBiosGhc old_env)
             , fromMaybe "" (lookup hieBiosGhcArgs old_env)
             )
-  output_file <- emptySystemTempFile "hie-bios"
+  output_file <- maybe (emptySystemTempFile "hie-bios") return
+                       (lookup hieBiosOutput old_env)
 
   -- Pipe stdout directly into the logger
   let process = (readProcessInDirectory work_dir fp args)
--- a/src/HIE/Bios/Cradle.hs
+++ b/src/HIE/Bios/Cradle.hs
@@ -49,7 +49,7 @@ import qualified Data.Conduit.Combinators as C
 import qualified Data.Conduit as C
 import qualified Data.Conduit.Text as C
 import qualified Data.Text as T
-import           Data.Maybe (fromMaybe)
+import           Data.Maybe (isNothing, fromMaybe)
 import           GHC.Fingerprint (fingerprintString)
 ----------------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


mpickering:
This will leak bios-output files into TMP. You need to clean up the temporary file after the cradle has finished running.

jneira:
@mpickering done! it is clearner now

src/HIE/Bios/Cradle.hs
```diff
@@ -637,33 +637,42 @@ readProcessWithOutputFile
...
+      -- Windows line endings are not converted so you have to filter out `'r` characters
+      loggingConduit = (C.decodeUtf8  C..| C.lines C..| C.filterE (/= '\r')  C..| C.map T.unpack C..| C.iterM l C..| C.sinkList)
+  (ex, stdo, stde) <- sourceProcessWithStreams process mempty loggingConduit loggingConduit
+  res <- withFile output_file ReadMode $ \handle -> do
+           hSetBuffering handle LineBuffering
+           !res <- force <$> hGetContents handle
+           return res
+
+  -- If the output was not provided by the user we delete it
+  when (isNothing mbHieBiosOut) (removeFile output_file)
```
fendor, line 668: Could this be turned into using the bracket idiom? Then it would be exception safe.

jneira: to ensure the removing of the file, but only if the file is temporal, created by hie-bios?

mpickering: @jneira I just fixed a related issue to this #179

src/HIE/Bios/Cradle.hs
```diff
@@ -637,33 +637,51 @@ readProcessWithOutputFile
...
+    let  loggingConduit = (C.decodeUtf8  C..| C.lines C..| C.filterE (/= '\r')  C..| C.map T.unpack C..| C.iterM l C..| C.sinkList)
     (ex, stdo, stde) <- sourceProcessWithStreams process mempty loggingConduit loggingConduit
-    !res <- force <$> hGetContents h
+    existsFile <- doesFileExist output_file
+    res <- if existsFile
+             then withFile output_file ReadMode $ \handle -> do
+                    hSetBuffering handle LineBuffering
+                    !res <- force <$> hGetContents handle
+                    return res
+             else return ""
```
mpickering, line 669: This should be an error rather than returning empty string?

src/HIE/Bios/Cradle.hs
```diff
@@ -637,33 +637,51 @@ readProcessWithOutputFile
...
+
     return (ex, stdo, stde, lines (filter (/= '\r') res))
 
     where
+      withHieBiosOutput :: [(String,String)] -> (FilePath -> IO a) -> IO a
+      withHieBiosOutput env action = do
+        let mbHieBiosOut = lookup hieBiosOutput env
+        case mbHieBiosOut of
+          Just file -> action file
```
mpickering, line 678: What if the filepath you pass in isn't an empty file alread?

src/HIE/Bios/Cradle.hs
```diff
@@ -637,33 +637,51 @@ readProcessWithOutputFile
...
+          Nothing -> withSystemTempDirectory "hie-bios" $
```
mpickering, line 679: Why create a directory rather than using `withSystemTempFile`?

jneira: cause it would cause the original file lock error in windows when the external process do a `whatever > %HIE_BIOS_OUTPUT%`  (`withSystemTempFile` open the file in read/write mode)

src/HIE/Bios/Cradle.hs
jneira: well, it seems there is test cases that actually dont write anything in the HIE_BIOS_OUTPUT (they failed before ad2a675 cause they dont even created the file to write in) so i assumed that it is a known-allowed case (see https://travis-ci.org/github/mpickering/hie-bios/jobs/683972142#L764 f.e.)
maybe is it the other way around and tests were wrong?

src/HIE/Bios/Cradle.hs
jneira: mmm, good catch, i did not take in account that case, will fix using the temp dir in that case too

src/HIE/Bios/Cradle.hs
jneira: well it turns out that `readProcessWithOutputFile` is used to make calls (well only one of them actually does it) that are not related with the args and consume the stdout/stderr:

https://github.com/mpickering/hie-bios/blob/f0abff9c855ea7e6624617df669825f3f62f723b/src/HIE/Bios/Cradle.hs#L489-L490

so it seems it is fine in some cases (one?) to ignore the fact that the call has not written anything to HIE_BIOS_OUTPUT

src/HIE/Bios/Cradle.hs
jneira: > What if the filepath you pass in isn't an empty file alread?

well, user is providing both the %HIE_BIOS_OUTPUT% and the external program that possibly modifies it so maybe there is a good reason to provide a non empty file (cause the external program only adds more info to it, f.e.)

src/HIE/Bios/Cradle.hs
mpickering: Can't you close the file again after it is opened by `withSystemTempFile`? That's what we do with the other temp files.

src/HIE/Bios/Cradle.hs
mpickering: I would really rather not return empty string here as imagine this situation.

* User consults the cradle, the file doesn't exist for some unknown reason
* The cradle reports "" as the output
* ghcide starts a session with no options
* User sees error about import of module failing

Now to get from the first step to the last step is not trivial to realise unless you are familar with how all the parts work.

src/HIE/Bios/Cradle.hs
jneira: mmm, i did not though in that alternative, will change it

src/HIE/Bios/Cradle.hs
jneira: ok, i've removed the file existence as you suggested check so it will crash early

jneira:
@mpickering @fendor i think it is done, thanks for the review and suggestions. 
I tested the last version in my win 10 with the ghc config and it works: it uses the file pointed by HIE_BIOS_OUTPUT and a temp file (like before) if it is not

mpickering:
Thanks LGTM. Over to you @fendor

jneira:
Thanks!, only a correction to my previous comments for the shake of being precise:  with the change from `withSystemTempDirectory` to `withSystemFileDirectory` the behaviour is the same as before the pr and the file is always created.
So the check for the existence of the file had no sense anymore (again like before the pr) and it is well removed.
However, the file can be empty and it would no trigger any error. I've checked it sending the output of `hadrian/hie-bios.bat` to `nul` and running `hie-bios check ghc\Main.hs` to see the temp file was created but compiler options are empty.
That is the reason that `stack path` call is succesful although it does not write anything to `$HIE_BIOS_OUTPUT`.

PR #179, opened 2020-05-05
Clean up hie-bios wrapper scripts after they are used

mpickering:


src/HIE/Bios/Cradle.hs
```diff
@@ -390,45 +390,53 @@ type GhcProc = (FilePath, [String])
...
+            (\loc h -> do
+                        hPutStr h cabalWrapper
+                        hClose h
+                        setMode loc
+                        k loc)
+
+  where
+    setMode wrapper_fp = do
+      setFileMode wrapper_fp accessModes
+      _check <- readFile wrapper_fp
```
fendor, line 422: Should we just delete that?

mpickering: I think we could, I had a look at where it was introduced and it used to print out the cradle to stderr after writing it to the file. That was modified when we started using a proper executable on windows to get the arguments.

PR #182, opened 2020-05-08
Remove unused environment variables to simplify code

### LINKED ISSUE #152: Reconsider use of environment variables  `HIE_BIOS_GHC` and  `HIE_BIOS_GHC_ARGS`
Nothing is currently using these env variables and they are undocumented because their usefulness is doubted.
The initial intention was to abstract `hie-bios` from the `ghc` executable on the path, but after the implementation we realized that it was not solving the problem.
So, currently it is an implementation artefact that no one is using and complicates the code quite a bit.

Either think of a use-case for the env variables `HIE_BIOS_GHC` and `HIE_BIOS_GHC_ARGS` or remove them entirely.

fendor:
Simplify the code, remove unused code-paths.

closes #152 
Nobody commented on this issue that these environment variables are currently used by anyone, the "feature" is undocumented and we dont use it in hie-bios at all. It just complicates the code.
Moreover, the intention of this "feature" was to fix a problem that was caused by something else, thus, delete them.

### DIFF
--- a/src/HIE/Bios/Cradle.hs
+++ b/src/HIE/Bios/Cradle.hs
@@ -323,7 +323,7 @@ biosWorkDir = findFileUpwards (".hie-bios" ==)
 biosDepsAction :: LoggingFunction -> FilePath -> Maybe FilePath -> IO [FilePath]
 biosDepsAction l wdir (Just biosDepsProg) = do
   biosDeps' <- canonicalizePath biosDepsProg
-  (ex, sout, serr, args) <- readProcessWithOutputFile l Nothing wdir biosDeps' []
+  (ex, sout, serr, args) <- readProcessWithOutputFile l wdir biosDeps' []
   case ex of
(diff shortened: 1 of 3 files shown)


mpickering:
Remind me again how this works with stack?

fendor:
Stack puts the correct "ghc" on the path within its process, so for `stack repl -w wrapper.sh` in `wrapper.sh` the correct `ghc` is on the PATH.

mpickering:
I approve this, but can't work out how to indicate that on GIthub.

fendor:
@jneira Yes, but I feel like this is the right approach, let the cradles make sure that they have the correct ghc set. In this particular case even the build-tools.
I agree, that it could be handy, but unless there is a particular use-case I would rather have this removed, as it isnt a well-desgined feature, but a monkey patch trying to fix something that never needed to be fixed.

PR #186, opened 2020-05-08
Store dependencies in CradleError

mpickering:
This allows you to attempt to rerun a failed cradle if any of the
dependencies change. It is not a very precise measure of why a cradle
failed but better than nothing.

Situations where this helps:

* You have a malformed `hie.yaml`
* You have a missing dependency in your cabal file
* You have a missing prefix in your hie.yaml
* You are missing a target in your cabal file

Situations where it doesn't help:

* A specific module in a dependency doesn't compile.

### DIFF
--- a/hie-bios.cabal
+++ b/hie-bios.cabal
@@ -183,6 +183,7 @@ test-suite bios-tests
       extra,
       tasty,
       tasty-hunit,
+      hspec-expectations,
       hie-bios,
       filepath,
       directory,
--- a/hie-bios.cabal
+++ b/hie-bios.cabal
@@ -52,6 +52,17 @@ Extra-Source-Files:     ChangeLog
                         tests/projects/multi-stack/hie.yaml
                         tests/projects/multi-stack/multi-stack.cabal
                         tests/projects/multi-stack/src/Lib.hs
+                        tests/projects/failing-bios/A.hs
+                        tests/projects/failing-bios/B.cabal
+                        tests/projects/failing-bios/hie.yaml
+                        tests/projects/failing-cabal/failing-cabal.cabal
+                        tests/projects/failing-cabal/hie.yaml
+                        tests/projects/failing-cabal/MyLib.hs
+                        tests/projects/failing-cabal/Setup.hs
+                        tests/projects/failing-stack/failing-stack.cabal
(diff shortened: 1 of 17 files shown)


src/HIE/Bios/Types.hs
```diff
@@ -73,16 +73,16 @@ data CradleLoadResult r
   deriving (Functor, Show)
 
 
-data CradleError = CradleError ExitCode [String] deriving (Show)
+data CradleError = CradleError [FilePath] ExitCode [String] deriving (Show)
```
fendor, line 76: This needs to be turned into a full record with documentation what the different fields denote

hie-bios.cabal
```diff
@@ -183,6 +194,8 @@ test-suite bios-tests
       extra,
       tasty,
       tasty-hunit,
+      tasty-expected-failure,
+      hspec-expectations,
```
fendor, line 198: error message isnt as nice yet as I hoped. We probably need to adapt the runner a bit.
Currently, the error message in a test-case failure looks like this (pretty formatted):
```
HUnitFailure
  (Just
     SrcLoc
       { srcLocPackage = "main"
       , srcLocModule = "Main"
       , srcLocFile = "tests/BiosTests.hs"
       , srcLocStartLine = 150
       , srcLocStartCol = 37
       , srcLocEndLine = 150
       , srcLocEndCol = 84
       })
  (Reason "Cradle loaded successfully")
```

src/HIE/Bios/Cradle.hs
```diff
@@ -118,8 +119,16 @@ addCradleDeps deps c =
...
       ca { runCradle = \l fp ->
-            (fmap (\(ComponentOptions os' dir ds) -> ComponentOptions os' dir (ds `union` deps)))
-              <$> runCradle ca l fp }
+            runCradle ca l fp
+              >>= \case
+                CradleSuccess (ComponentOptions os' dir ds) ->
+                  pure $ CradleSuccess (ComponentOptions os' dir (ds `union` deps))
+                CradleFail err ->
+                  pure $ CradleFail
+                    (err { cradleErrorDependencies = cradleErrorDependencies err `union` deps })
```
fendor, line 128: This had to be added

PR #188, opened 2020-05-21
add shell and dependency-shell attributes to bios cradle type

### LINKED ISSUE #158: Support arguments to bios process
My bios script is just

```bash
#!/usr/bin/env bash
ob internal export-some-stuff > "$HIE_BIOS_OUTPUT"
```

It would be nicer if I could just inline this into `hie.yaml` directly with something like

```yaml
cradle: { bios: { program: 'bash -c "ob internal export-some-stuff > \"$HIE_BIOS_OUTPUT\""' } }
```

hyperfekt:
This changes `readProcessWithOutputFile` to take a `createProcess` datatype instead of a program with arguments so the bios cradle type can pass in a shell command.
I tried to avoid additional imports. Please feel free to make changes as you see fit, ask questions or give feedback.

Fixes #158.


fendor:
Awesome! 
I am interested on how well this works on windows. If this works on windows, can we maybe get a test for windows as well?

hyperfekt:
The project test I added is intended to work on Windows, too, but I have no way to actually run it.

jneira:
Hi! first of all, great work, the pr looks good but not sure if the test case is correct: in windows shells you have to refer to en vars with `%HIE_BIOS_OUTPUT%`, not `$HIE_BIOS_OUTPUT`.
So i am not sure if the test had to be succesful in windows.
Consider the shell session:
```
PS D:\dev\ws\haskell\hie-bios\tests\projects\simple-bios-shell> dir -name
A.hs
B.hs
hie.yaml
PS D:\dev\ws\haskell\hie-bios\tests\projects\simple-bios-shell> hie-bios debug .\A.hs    
Root directory:      D:\dev\ws\haskell\hie-bios\tests\projects\simple-bios-shell
Component directory: D:\dev\ws\haskell\hie-bios\tests\projects\simple-bios-shell
GHC options:
System libraries:    D:\bin\stack\x86_64-windows\ghc-8.8.3\lib
Config Location:     D:\dev\ws\haskell\hie-bios\tests\projects\simple-bios-shell\hie.yaml
Cradle:              Cradle {cradleRootDir = "D:\\dev\\ws\\haskell\\hie-bios\\tests\\projects\\simple-bios-shell", cradleOptsProg = CradleAction: Bios}
Dependencies:
PS D:\dev\ws\haskell\hie-bios\tests\projects\simple-bios-shell> dir -name
$HIE_BIOS_OUTPUT
A.hs
B.hs
hie.yaml
PS D:\dev\ws\haskell\hie-bios\tests\projects\simple-bios-shell> cat `$HIE_BIOS_OUTPUT
"-Wall"  
"A"  
"B"
```

Simply changing to the correct dos syntax make it work, so the code is correct:

```
PS D:\dev\ws\haskell\hie-bios\tests\projects\simple-bios-shell> cat .\hie.yaml
cradle:
  bios: {shell: 'echo "-Wall" >> %HIE_BIOS_OUTPUT% && echo "A" >> %HIE_BIOS_OUTPUT% && echo "B" >> %HIE_BIOS_OUTPUT%' }      
PS D:\dev\ws\haskell\hie-bios\tests\projects\simple-bios-shell> hie-bios debug .\A.hs   
Root directory:      D:\dev\ws\haskell\hie-bios\tests\projects\simple-bios-shell
Component directory: D:\dev\ws\haskell\hie-bios\tests\projects\simple-bios-shell
GHC options:         ""-Wall"  " ""A"  " ""B" "
System libraries:    D:\bin\stack\x86_64-windows\ghc-8.8.3\lib
Config Location:     D:\dev\ws\haskell\hie-bios\tests\projects\simple-bios-shell\hie.yaml
Cradle:              Cradle {cradleRootDir = "D:\\dev\\ws\\haskell\\hie-bios\\tests\\projects\\simple-bios-shell", cradleOptsProg = CradleAction: Bios}
Dependencies:
```

But i am a liitle bit worried by the fact that the test should have failed

jneira:
And it shows one of the shortcomings of the cradle: in general, you will need one version for each type of shell you want to support.
But it is not related with the pr 😄

hyperfekt:
This little trick should allow using both Windows-specific and POSIX commands in one command string. I have no idea why the test was passing though or even what exactly its conditions are, since I just blindly adapted it from the others. Maybe the project tests aren't really as thorough as thought?

jneira:
@hyperfekt i think those tests dont check the compiler args returned by the cradle, only that loading the cradle returns `CradleSuccess` (maybe @fendor could confirm that):

https://github.com/mpickering/hie-bios/blob/6ab59efaa40438c96dc49549d4ea5e6a86e0aedf/tests/BiosTests.hs#L101-L110

In that case i would not worry much about.

src/HIE/Bios/Config.hs
```diff
@@ -127,16 +130,28 @@ parseCabal = parseStackOrCabal Cabal CabalMulti
 parseBios :: Value -> Parser (CradleType a)
 parseBios (Object x)
     | 2 == Map.size x
-    , Just (String biosProgram) <- Map.lookup "program" x
-    , Just (String biosDepsProgram) <- Map.lookup "dependency-program" x
-    = return $ Bios (T.unpack biosProgram) (Just (T.unpack biosDepsProgram))
+    , Just biosCallable <- exclusive (stringTypeFromMap Program "program") (stringTypeFromMap Command "shell")
+    , Just biosDepsCallable <- exclusive (stringTypeFromMap Program "dependency-program") (stringTypeFromMap Command "dependency-shell")
```
mpickering, line 134: Looks like here you could have `program` and `dependency-shell` which I suppose is acceptable?

src/HIE/Bios/Cradle.hs
```diff
@@ -487,11 +497,11 @@ stackAction work_dir mc l _fp = do
   -- Same wrapper works as with cabal
   withCabalWrapperTool ghcProcArgs work_dir $ \wrapper_fp -> do
     (ex1, _stdo, stde, args) <-
-      readProcessWithOutputFile l work_dir
-              "stack" $ ["repl", "--no-nix-pure", "--with-ghc", wrapper_fp]
-                        ++ [ comp | Just comp <- [mc] ]
+      readProcessWithOutputFile l $ (proc "stack" (["repl", "--no-nix-pure", "--with-ghc", wrapper_fp]
```
mpickering, line 500: Can you add back the directory argument to `readProcessWithOutputFile` and set it inside this function rather than at all the call sites?

src/HIE/Bios/Cradle.hs
```diff
@@ -623,30 +633,24 @@ findFile p dir = do
...
--- * The path to the GHC version to use is supplied in the environment variable @HIE_BIOS_GHC@.
---   Additionally, arguments to ghc are supplied via @HIE_BIOS_GHC_ARGS@
 readProcessWithOutputFile
   :: LoggingFunction -- ^ Output of the process is streamed into this function.
-  -> FilePath -- ^ Working directory. Process is executed in this directory.
-  -> FilePath -- ^ Process to call.
-  -> [String] -- ^ Arguments to the process.
+  -> CreateProcess
   -> IO (ExitCode, [String], [String], [String])
-readProcessWithOutputFile l work_dir fp args = do
```
mpickering, line 641: Add back `work_dir` argument I think?

mpickering:
Thanks, very nice patch. Just a couple of comments.

src/HIE/Bios/Config.hs
hyperfekt: Yes, that was intentional. I didn't see a good reason to disallow it, and I added a test and documentation explicitly for this kind of mixing.

src/HIE/Bios/Cradle.hs
hyperfekt: Agreed, good point.

PR #197, opened 2020-06-11
Improve filtering of rts arguments from stack and cabal cradles

fendor:
Handles the case that the rts flags produced by the cradles are
within a single string.
Before we only allowed the rts flags to occur in the following form:

> ["+RTS", "-H32m","-RTS"]

and now we allow:

> ["+RTS -H32m -RTS"]

and filter the ComponentOptions appropriately.

### DIFF
--- a/src/HIE/Bios/Cradle.hs
+++ b/src/HIE/Bios/Cradle.hs
@@ -217,7 +217,7 @@ defaultCradle cur_dir =
     }
 
 ---------------------------------------------------------------
--- The none cradle tells us not to even attempt to load a certain directory
+-- | The none cradle tells us not to even attempt to load a certain directory
 
 noneCradle :: FilePath -> Cradle a
 noneCradle cur_dir =
@@ -230,7 +230,7 @@ noneCradle cur_dir =
     }
 
 ---------------------------------------------------------------
--- The multi cradle selects a cradle based on the filepath
+-- | The multi cradle selects a cradle based on the filepath
 
 multiCradle :: (b -> Cradle a) -> FilePath -> [(FilePath, CradleConfig b)] -> Cradle a
 multiCradle buildCustomCradle cur_dir cs =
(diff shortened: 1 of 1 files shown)


src/HIE/Bios/Cradle.hs
```diff
@@ -457,13 +456,14 @@ cabalAction work_dir mc l fp = do
...
-removeRTS ("+RTS" : xs)  =
-  case dropWhile (/= "-RTS") xs of
-    [] -> []
-    (_ : ys) -> removeRTS ys
-removeRTS (y:ys)         = y : removeRTS ys
+removeRTS (y:ys)
+  -- rts flags are spanning multiple entries
+  | y == "+RTS" = removeRTS (tail $ dropWhile (/= "-RTS") ys)
+  -- rts flags are in a single string
+  | "+RTS" `isPrefixOf` y && "-RTS" `isSuffixOf` y = removeRTS ys
```
rjmk, line 465: I'm not sure that we should be checking the suffix. I don't know what to do in the case that `ys` is non-empty and `"-RTS"` isn't a suffix, but if `ys` is empty, we shouldn't need the `"-RTS"` suffix.

fendor: Thank you for your review! 

I think it is kind of debatable what to do.
Theoretically, it is possible to have a `+RTS` without a `-RTS` if it is the last argument. 
But if there are flags such as `["+RTS -H32m", "-RTS"]` I think we should just let it crash unless we see that we need to fix it.

src/HIE/Bios/Cradle.hs
```diff
@@ -457,13 +456,14 @@ cabalAction work_dir mc l fp = do
 removeInteractive :: [String] -> [String]
 removeInteractive = filter (/= "--interactive")
 
--- Strip out any ["+RTS", ..., "-RTS"] sequences in the command string list.
+-- | Strip out any ["+RTS", ..., "-RTS"] sequences in the command string list.
 removeRTS :: [String] -> [String]
```
rjmk, line 460:
```suggestion
data InRTS = OutsideRTS | InsideRTS

removeRTS :: [String] -> [String]
removeRTS = go OutsideRTS where
  go :: InRTS -> [String] -> [String]
  go _ [] = []
  go OutsideRTS (y:ys)
    | "+RTS" `isPrefixOf` y = removeRTS (if "-RTS" `isSuffixOf y then OutsideRTS else InsideRTS) ys
    | otherwise = y : removeRTS ys
  go InsideRTS (y:ys) = removeRTS (if y = "-RTS" then OutsideRTS else InsideRTS) ys
```

src/HIE/Bios/Cradle.hs
rjmk: Ah, sorry I didn't see your comment before my suggestion. Makes sense!

PR #205, opened 2020-06-14
Fix absolute mains

fendor:
Adds failing test-case.

### DIFF
--- a/src/HIE/Bios/Environment.hs
+++ b/src/HIE/Bios/Environment.hs
@@ -1,5 +1,5 @@
 {-# LANGUAGE RecordWildCards, CPP #-}
-module HIE.Bios.Environment (initSession, getRuntimeGhcLibDir, getRuntimeGhcVersion, makeDynFlagsAbsolute, getCacheDir, addCmdOpts) where
+module HIE.Bios.Environment (initSession, getRuntimeGhcLibDir, getRuntimeGhcVersion, makeDynFlagsAbsolute, makeTargetsAbsolute, getCacheDir, addCmdOpts) where
 
 import CoreMonad (liftIO)
 import GHC (GhcMonad)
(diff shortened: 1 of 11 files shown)


fendor:
@maksbotan Could you check if this fixes your problem, too?

maksbotan:
Looks like it broke something :(
```
    space stack:               FAIL (1.56s)
      Finding Cradle for: /tmp/dist-test.rmDk/hie-bios-0.5.0/tests/projects/space stack/B.hs
      Loading Cradle: Just "/tmp/dist-test.rmDk/hie-bios-0.5.0/tests/projects/space stack/hie.yaml"
      Initialise Flags                                               (1.56s)
      Initial module load
        can't find file: /tmp/dist-test.rmDk/hie-bios-0.5.0/tests/projects/space/B.hs
```

maksbotan:
On the other hand, it does fix my issue, thanks!

I like you code more, I just didn't think about looking into what `Target` is :)

fendor:
@maksbotan It actually uncovered an existing bug #206 which needs to be fixed.

src/HIE/Bios/Environment.hs
```diff
@@ -47,9 +47,20 @@ initSession  ComponentOptions {..} = do
...
+    return targets'
+
+----------------------------------------------------------------
+
+normaliseTargets :: FilePath -> [G.Target] -> [G.Target]
+normaliseTargets wdir = map (\target -> target {G.targetId = normaliseTargetId wdir (G.targetId target)})
+
+normaliseTargetId :: FilePath -> G.TargetId -> G.TargetId
+normaliseTargetId wdir (G.TargetFile fp phase) = G.TargetFile (prependIfRelative wdir fp) phase
+normaliseTargetId _ tid = tid
```
fendor, line 63: This should probably be named 'makeTargetsAbsolute' to mirror the naming of `makeDynFlagsAbsolute`.
@jneira, @bubba opinions?

bubba: Yeah normalise to me means something more like `canonicalizePath`

PR #206, opened 2020-06-14
Fix printing of current directory in wrapper script

fendor:
Noticed with current hie-bios:
```bash
$ hie-bios debug A.hs
Root directory:      /home/munin/Documents/haskell/hie-bios/tests/projects/space stack
Component directory: /home/munin/Documents/haskell/hie-bios/tests/projects/space
Dependencies:        stackproj.cabal package.yaml stack.yaml
```

Notice the root and component directory. This needs an additional test.

Seems to be our first unix exclusive bug :confetti_ball:

### DIFF
--- a/wrappers/cabal
+++ b/wrappers/cabal
@@ -5,7 +5,7 @@ function out(){
 }
 
 if [ "$1" == "--interactive" ]; then
-  out $(pwd)
+  out "$(pwd)"
   for arg in "$@"; do
     out "$arg"
   done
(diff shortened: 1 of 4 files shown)


wrappers/cabal
```diff
@@ -5,7 +5,7 @@ function out(){
 }
 
 if [ "$1" == "--interactive" ]; then
-  out $(pwd)
+  out "$(pwd)"
```
maksbotan, line 8: Haha, nice! Gotta love strings.

fendor: Or gotta hate bash >_>

PR #207, opened 2020-06-14
Add getRuntimeGhcLibDir and getRuntimeGhcVersion, via runGhc

bubba:
This PR moves the logic in ghcide/hls of getting the **runtime** ghc library directory and version into hie-bios, and in doing so moves the responsibility to the cradle

### DIFF
--- a/src/HIE/Bios/Cradle.hs
+++ b/src/HIE/Bios/Cradle.hs
@@ -40,9 +40,7 @@ import System.Environment
 import Control.Applicative ((<|>), optional)
 import System.IO.Temp
 import System.IO.Error (isPermissionError)
-import Data.Char
 import Data.List
-import Data.Foldable
 import Data.Ord (Down(..))
 
 import System.PosixCompat.Files
@@ -217,7 +215,7 @@ defaultCradle cur_dir =
         { actionName = Types.Default
         , runCradle = \_ _ ->
             return (CradleSuccess (ComponentOptions [] cur_dir []))
-        , runGhcLibDir = getGhcOnPathLibDir cur_dir
+        , runGhc = runGhcOnPath cur_dir
         }
     }
 
@@ -231,7 +229,7 @@ noneCradle cur_dir =
     , cradleOptsProg = CradleAction
         { actionName = Types.None
         , runCradle = \_ _ -> return CradleNone
-        , runGhcLibDir = pure Nothing
+        , runGhc = const $ pure Nothing
         }
     }
 
@@ -245,13 +243,13 @@ multiCradle buildCustomCradle cur_dir cs =
     , cradleOptsProg = CradleAction
         { actionName = multiActionName
         , runCradle  = \l fp -> canonicalizePath fp >>= multiAction buildCustomCradle cur_dir cs l
-        , runGhcLibDir =
+        , runGhc = \args ->
             -- We're being lazy here and just returning the ghc path for the
             -- first non-none cradle. This shouldn't matter in practice: all
             -- sub cradles should be using the same ghc version!
             case filter (not . isNoneCradleConfig) $ map snd cs of
               [] -> return Nothing
-              (cfg:_) -> runGhcLibDir $ cradleOptsProg $
+              (cfg:_) -> flip runGhc args $ cradleOptsProg $
                 getCradle buildCustomCradle (cfg, cur_dir)
         }
     }
@@ -326,7 +324,7 @@ directCradle wdir args =
         { actionName = Types.Direct
         , runCradle = \_ _ ->
             return (CradleSuccess (ComponentOptions args wdir []))
-        , runGhcLibDir = getGhcOnPathLibDir wdir
+        , runGhc = runGhcOnPath wdir
         }
     }
 
@@ -342,7 +340,7 @@ biosCradle wdir biosCall biosDepsCall =
     , cradleOptsProg   = CradleAction
         { actionName = Types.Bios
         , runCradle = biosAction wdir biosCall biosDepsCall
-        , runGhcLibDir = getGhcOnPathLibDir wdir
+        , runGhc = runGhcOnPath wdir
         }
     }
 
@@ -394,7 +392,7 @@ cabalCradle wdir mc =
     , cradleOptsProg   = CradleAction
         { actionName = Types.Cabal
         , runCradle = cabalAction wdir mc
-        , runGhcLibDir = optional $ fmap trim $ do
+        , runGhc = \args -> optional $ do
             -- Workaround for a cabal-install bug on 3.0.0.0:
             -- ./dist-newstyle/tmp/environment.-24811: createDirectory: does not exist (No such file or directory)
             -- (It's ok to pass 'dist-newstyle' here, as it can only be changed
@@ -402,7 +400,8 @@ cabalCradle wdir mc =
             -- using in our call to v2-exec)
             createDirectoryIfMissing True (wdir </> "dist-newstyle" </> "tmp")
             -- Need to pass -v0 otherwise we get "resolving dependencies..."
-            readProcessWithCwd wdir "cabal" ["v2-exec", "ghc", "-v0", "--", "--print-libdir"] ""
+            readProcessWithCwd
+              wdir "cabal" (["v2-exec", "ghc", "-v0", "--"] ++ args) ""
         }
     }
 
@@ -540,8 +539,9 @@ stackCradle wdir mc =
     , cradleOptsProg   = CradleAction
         { actionName = Types.Stack
         , runCradle = stackAction wdir mc
-        , runGhcLibDir = optional $ fmap trim $
-            readProcessWithCwd wdir "stack" ["exec", "--silent", "ghc", "--", "--print-libdir"] ""
+        , runGhc = \args -> optional $
+            readProcessWithCwd
+              wdir "stack" (["exec", "--silent", "ghc", "--"] <> args) ""
         }
     }
 
@@ -728,8 +728,8 @@ readProcessWithOutputFile l work_dir cp = do
 
     where
       withHieBiosOutput :: [(String,String)] -> (FilePath -> IO a) -> IO a
-      withHieBiosOutput env action = do
-        let mbHieBiosOut = lookup hieBiosOutput env
+      withHieBiosOutput env' action = do
+        let mbHieBiosOut = lookup hieBiosOutput env'
         case mbHieBiosOut of
           Just file@(_:_) -> action file
           _ -> withSystemTempFile "hie-bios" $
@@ -745,14 +745,9 @@ makeCradleResult (ex, err, componentDir, gopts) deps =
         let compOpts = ComponentOptions gopts componentDir deps
         in CradleSuccess compOpts
 
--- Used for clipping the trailing newlines on some commands
-trim :: String -> String
-trim = dropWhileEnd isSpace
-
 -- | Calls @ghc --print-libdir@, with just whatever's on the PATH.
-getGhcOnPathLibDir :: FilePath -> IO (Maybe FilePath)
-getGhcOnPathLibDir wdir = optional $
-  fmap trim $ readProcessWithCwd wdir "ghc" ["--print-libdir"] ""
+runGhcOnPath :: FilePath -> [String] -> IO (Maybe String)
+runGhcOnPath wdir args = optional $ readProcessWithCwd wdir "ghc" args ""
 
 -- | Wrapper around 'readCreateProcess' that sets the working directory
 readProcessWithCwd :: FilePath -> FilePath -> [String] -> String -> IO String
--- a/src/HIE/Bios/Cradle.hs
+++ b/src/HIE/Bios/Cradle.hs
@@ -37,10 +37,12 @@ import Control.Monad
 import System.Info.Extra
 import Control.Monad.IO.Class
 import System.Environment
-import Control.Applicative ((<|>))
+import Control.Applicative ((<|>), optional)
 import System.IO.Temp
 import System.IO.Error (isPermissionError)
+import Data.Char
 import Data.List
+import Data.Foldable
 import Data.Ord (Down(..))
 
 import System.PosixCompat.Files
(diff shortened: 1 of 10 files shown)


bubba:
Closes https://github.com/mpickering/hie-bios/issues/151

tests/BiosTests.hs
```diff
@@ -4,9 +4,11 @@
 module Main where
 
 import Test.Tasty
-import Test.Tasty.ExpectedFailure
 import Test.Tasty.HUnit
 import Test.Hspec.Expectations
+#if __GLASGOW_HASKELL__ < 810
+import Test.Tasty.ExpectedFailure
+#endif
```
fendor, line 11: Why does this change make sense?

src/HIE/Bios/Cradle.hs
```diff
@@ -378,6 +393,9 @@ cabalCradle wdir mc =
     , cradleOptsProg   = CradleAction
         { actionName = Types.Cabal
         , runCradle = cabalAction wdir mc
+        , getGhcPath = optional $ fmap trim $
+            -- Need to pass -v0 otherwise we get "resolving dependencies..."
+            readProcess "cabal" ["v2-exec", "which", "-v0", "--", "ghc"] ""
```
fendor, line 398: Does this work for completely uninitialized cabal projects?
Some tests would be nice

src/HIE/Bios/Cradle.hs
```diff
@@ -239,6 +243,14 @@ multiCradle buildCustomCradle cur_dir cs =
     , cradleOptsProg = CradleAction
         { actionName = multiActionName
         , runCradle  = \l fp -> canonicalizePath fp >>= multiAction buildCustomCradle cur_dir cs l
+        , getGhcPath =
+            -- We're being lazy here and just returning the ghc path for the
+            -- first non-none cradle. This shouldn't matter in practice: all
+            -- sub cradles should be using the same ghc version!
```
fendor, line 249: This should probably be "must"

src/HIE/Bios/Environment.hs
```diff
@@ -53,13 +56,22 @@ initSession  ComponentOptions {..} = do
...
+-- | Obtain the directory for the ghc library directory.
+-- It tries this specific order of paths:
+-- 1. the NIX_GHC_LIBDIR if it is set
+-- 2. calling --print-libdir with the cradle's ghc
+-- 3. using ghc-paths
+-- Note that 3 will only work if we are running on the same machine that
+-- hie-bios compiled on, so we avoid using it unless we have to!
+getGhcLibDir :: Cradle a -> IO FilePath
+getGhcLibDir cradle =
+  fmap (fromMaybe Paths.libdir) $ runMaybeT $ fromNix <|> fromCradle
```
fendor, line 68: Do we know what should take precedence? 
In my experience with nix, ghc is wrapped to give the correct libdir.
The order should be in my opinion: Cradle Ghc -> Nix env var -> Paths.libdir

src/HIE/Bios/Ghc/Api.hs
```diff
@@ -32,29 +31,25 @@ import HIE.Bios.Flags
 ----------------------------------------------------------------
 
 -- | Converting the 'Ghc' monad to the 'IO' monad. All exceptions are ignored and logged.
-withGHC :: FilePath  -- ^ A target file displayed in an error message.
+withGHC :: Cradle c  -- ^ The cradle to use for resolving the lib dir
```
fendor, line 34: Not entirely sure I like this API. The ghc session is not dependent on the cradle, only on the libdir.
To me, this usage sends a signal that the cradle is somehow used in the session and needs to be known before launching the session. (my opinion is debatable here)
(btw, neither ghcide/hls nor haskell-ide-engine uses it)

src/HIE/Bios/Ghc/Check.hs
```diff
@@ -53,22 +54,23 @@ checkSyntax cradle files = withGhcT $ do
...
 dup :: a -> (a, a)
 dup x = (x, x)
 
 ----------------------------------------------------------------
 
 -- | Set 'DynFlags' equivalent to "-Wall".
-setAllWarningFlags :: DynFlags -> DynFlags
-setAllWarningFlags df = df { warningFlags = allWarningFlags }
+setAllWarningFlags :: FilePath -> DynFlags -> DynFlags
+setAllWarningFlags topDir df = df { warningFlags = allWarningFlags topDir }
```
fendor, line 68: this should rather "libdir"?

src/HIE/Bios/Ghc/Check.hs
```diff
@@ -53,22 +54,23 @@ checkSyntax cradle files = withGhcT $ do
...
 
 {-# NOINLINE allWarningFlags #-}
-allWarningFlags :: Gap.WarnFlags
-allWarningFlags = unsafePerformIO $ do
-    mlibdir <- getSystemLibDir
-    G.runGhcT mlibdir $ do
+allWarningFlags :: FilePath -> Gap.WarnFlags
+allWarningFlags topDir = unsafePerformIO $
```
fendor, line 72: same here, isnt this the libdir?

src/HIE/Bios/Cradle.hs
jneira: Yeah, i think that line would fail in windows
```
PS D:\dev\ws\haskell\hls> cabal v2-exec which -v0 -- ghc
cabal.exe: The program 'which' is required but it could not be found.
```

src/HIE/Bios/Cradle.hs
jneira: Not sure about that, the multi combinations could give us as first ghc a wrong one. F.e. if you use a direct cradle it can give us the ghc on path but stack or cabal the ghc configured for the project.
However it is probable that the direct cradle could work with the build tool one but not the other way around.
Would be not feasible return a list of unique ghc's and let the client code pick the first one or handle the list in other way?  i can imagine you could choose the newer ghc, f.e.

src/HIE/Bios/Cradle.hs
```diff
@@ -495,6 +513,8 @@ stackCradle wdir mc =
     , cradleOptsProg   = CradleAction
         { actionName = Types.Stack
         , runCradle = stackAction wdir mc
+        , getGhcPath = optional $ fmap trim $
+            readProcess "stack" ["exec", "which", "--silent", "--", "ghc"] ""
```
jneira, line 517: This may not be portable neither:
```
PS D:\dev\ws\haskell\hls> stack exec which --silent -- ghc
/d/bin/stack/x86_64-windows/ghc-8.6.5/bin/ghc
```
stack uses `which` from its msys installation but the path has the msys format, not sure if it could be usable as is.

What about

```
D:\dev\ws\haskell\hls>stack path --compiler-exe
D:\bin\stack\x86_64-windows\ghc-8.6.5\bin\ghc-8.6.5.exe
```

src/HIE/Bios/Environment.hs
jneira: As ghc-paths will be wrong in the general case maybe we should remove it and returns `IO (Maybe Filepath)` to make clear we cant return it for all cases in a reliable way and force the client code handle that possibility (maybe using ghc-paths at the end!)

src/HIE/Bios/Cradle.hs
```diff
@@ -326,6 +340,7 @@ biosCradle wdir biosCall biosDepsCall =
     , cradleOptsProg   = CradleAction
         { actionName = Types.Bios
         , runCradle = biosAction wdir biosCall biosDepsCall
+        , getGhcPath = findExecutable "ghc"
```
jneira, line 343: Afaik the dependencies of the bios cradle is a responsability of the bios program (see biosDeps): should not be the ghc used by that bios program returned by itself? Or at least being able to take in account it and fallback to the ghc on path.

jneira:
@bubba thanks a lot for this, not only it is necessary for the demanded precompiled ide binaries but i think it is a great feature for hie-bios standalone.
As @fendor noted we have to be specially careful, sorry i have been too fuzzy 😅

tests/BiosTests.hs
bubba: Fixes a redundant import warning on ghc 8.10

src/HIE/Bios/Ghc/Check.hs
bubba: The GHC api seems to refer to the library directory as topDir, that's the function I'm using to get it, but I'm happy out to keep it as libDir for consistency. The true bikeshed question though is, is it libDir or libdir? 🤔

src/HIE/Bios/Ghc/Api.hs
bubba: Ok, I'm happy enough to change this to the libdir, and then the users of this API can just run the cradle themselves to extract it

src/HIE/Bios/Environment.hs
bubba: My aim here was to mirror what currently happens inside ghcide/hls, and I presume this nix env var might be used for the direct cradles where it's calling `ghc` without cabal or stack: https://github.com/digital-asset/ghcide/blob/9e984b56cfb62a83829975210ce3f14c668912f6/docs/Setup.md#issues-with-nix

bubba: If the client code is just going to end up calling ghc-paths then shouldn't we just do it here as well? We might want to rename `getGhcLibDir` to something more honest though like `mostlyAccurateGuessGhcLibDir`, since the *only* scenario where ghc-paths doesn't work will be when moving binaries across machines. On the other hand, if we are using static binaries then we would really want to error here instead of silently giving the wrong path. But then we would end up splitting the libdir logic even further back into the clients

src/HIE/Bios/Ghc/Api.hs
bubba: Come to think of it, if we make the user pass in the libdir then there's no point really in having `withGHC`/`withGhcT`, since all it did was handle the libdir part

bubba: I'd also advocate for just removing withGHC as well, since it's not used by ghcide/hls and is a carryover from ghc-mod

src/HIE/Bios/Environment.hs
jneira: Well i suppose client code could do a final sanity check executing the ghc returned by the cradle (`--numeric-version` ?) and catch the error. Not elegant but doable

src/HIE/Bios/Cradle.hs
bubba: Is there a way for a bios program to return the path to the GHC it intends to use? From my understanding it just returns newline separated arguments in stdout

fendor: It may still want define the libdir, right?

src/HIE/Bios/Ghc/Api.hs
fendor: Yeah, remove it.

src/HIE/Bios/Cradle.hs
bubba: The ghc libdir should be the same ghc version for each sub-cradle though. This would mean it doesn't matter which one is used to load the ghc session
The only case where it would make a difference  would be if two sub-cradles actually had different ghc versions, and then returned two different ghc libdirs of different ghc versions.
But then in that case ghcide/hls wouldn't be able to load them both anyway!

I guess we could return a list anyway for ghcide/hls and they could just called `head`. But I think it would be nice to handle as much of this stuff in hie-bios as possible

src/HIE/Bios/Cradle.hs
```diff
@@ -2,6 +2,7 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TupleSections #-}
 {-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE CPP #-}
```
bubba, line 5: Please forgive me

src/HIE/Bios/Environment.hs
bubba: @jneira `getGhcOnPathLibDir` should now do a check of the ghc version beforehand now

src/HIE/Bios/Cradle.hs
bubba: I'm not sure if there will be a good way to have the bios program pass the library directory back, seeing as programs are only really aware of filepaths being passed to them, either through the shell arguments or by the `HIE_BIOS_ARG` environment variable. Also I can't really imagine how the bios programs are going to return the library directory in the first place. 
With that said, the current way haskell-language-server works out the project version is by calling `ghc` on the path anyway, and this should fall back to the hardcoded ghc-paths should that fail. So I don't think there would be much of a difference.

jneira: Well, an option could be add another optional config option like the existing one for dependencies:
```
cradle:
  bios:
    program: ./flags.sh
    dependency-program: ./dependency.sh
    ghc-path-program: ./ghc-path.sh
```
But i am not aware of a bios program that needs it so i would merge this without it and add it on concrete user request afterwards

src/HIE/Bios/Cradle.hs
jneira: Uh i completly missed that comment, sorry
Actually i am not sure if return a list without knowing which cradle is the source of each ghc version. I dont have an alternative that it is simple, maybe return an error or warning if not all ghcs are he same version? 🤔 

If you think that this could be merged as is, it will be fine for me too

bubba:
For posterity, the issue with hie-bios failing on ghc was due to the sanity check that the ghc versions matched when fetching the library directory for GHC on the path. We just removed this in the end, since we now have a separate function for checking the project version.

And upon further reflection, I think hie-bios shouldn't try to act smart here – the runtime ghc version for a bios or plain/default cradle should *always* just consult `ghc` on the PATH. Not `ghc-8.x.y` with whatever version hie-bios was compiled with.
This will only get confusing, and the version of what hie-bios was compiled with really shouldn't determine the version of ghc selected since we will be using this logic in `haskell-langugage-server-wrapper`, which is compiled with an arbitrary ghc version!

jneira:
> Not ghc-8.x.y with whatever version hie-bios was compiled with.

Yeah, moreover now it is named `getRuntimeGhcLibDir`, i would remove the fromGhcPaths alternative

bubba:
Was trying to figure out why appveyor was failing with stack and ghc 8.10.1 so I set up GitHub CI to see if would make a difference but it didn't. I think there's something funny going on with that combination anyway so I've just commented out those tests again. I've left in the GitHub CI anyway, should be handy

PR #209, opened 2020-06-21
Make Cradle dependencies for stack and cabal more reasonable

### LINKED ISSUE #203: Rework paths in Cradle Dependencies
Original issue: https://github.com/digital-asset/ghcide/issues/633

The dependencies could be relative to the `cradleRootDir` or to `componentRootDir` e.g.
```
.
├── cabal.project
├── servant-server
│   └── servant-server.cabal
└─── servant-docs
    └── servant-docs.cabal
```

```
Options for 'servant-server'
Root directory:      ./servant
Component directory: ./servant/servant-server
Dependencies:        stack.yaml servant-server.cabal package.yaml
```

```
Options for 'servant-docs'
Root directory:      ./servant
Component directory: ./servant/servant-docs
Dependencies:        stack.yaml servant-docs.cabal package.yaml
```

We need to reload both components  in case of change to `./servant/stack.yaml` but only one for changes to `./servant/servant-server/servant-server.cabal`, `./servant/servant-server/package.yaml` and `./servant/servant-docs/servant-docs.cabal`, `./servant/servant-docs/package.yaml` respectively.

A proposed solution:
Change the API of cradle dependencies in hie-bios to give absolute paths since this is the only location where we can know this. (Another breaking change, yay)
Unfortunately, this does not play nicely with the latest breaking change to extend a CradleError with the dependencies of the component, since we dont know the component dir in that case :(

fendor:
Closes #203

### DIFF
--- a/src/HIE/Bios/Cradle.hs
+++ b/src/HIE/Bios/Cradle.hs
@@ -405,11 +405,25 @@ cabalCradle wdir mc =
         }
     }
 
-cabalCradleDependencies :: FilePath -> IO [FilePath]
-cabalCradleDependencies rootDir = do
-    cabalFiles <- findCabalFiles rootDir
-    return $ cabalFiles ++ ["cabal.project", "cabal.project.local"]
+-- | @'cabalCradleDependencies' rootDir componentDir@.
+-- Compute the dependencies of the cabal cradle based
+-- on the cradle root and the component directory.
+--
+-- Directory 'componentDir' is a sub-directory where we look for
+-- package specific cradle dependencies, such as a '.cabal' file.
+--
+-- Found dependencies are relative to 'rootDir'.
+cabalCradleDependencies :: FilePath -> FilePath -> IO [FilePath]
+cabalCradleDependencies rootDir componentDir = do
+    let relFp = makeRelative rootDir componentDir
+    cabalFiles' <- findCabalFiles componentDir
+    let cabalFiles = map (relFp </>) cabalFiles'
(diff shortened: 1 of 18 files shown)


src/HIE/Bios/Cradle.hs
```diff
@@ -381,11 +381,25 @@ cabalCradle wdir mc =
...
+-- Directory 'componentDir' is a sub-directory where we look for
+-- package specific cradle dependencies, such as a '.cabal' file.
+--
+-- Found dependencies are relative to 'rootDir'.
+cabalCradleDependencies :: FilePath -> FilePath -> IO [FilePath]
+cabalCradleDependencies rootDir componentDir = do
+    let relFp = makeRelative rootDir componentDir
+    cabalFiles' <- findCabalFiles componentDir
+    let cabalFiles = map (relFp </>) cabalFiles'
+    return $ map normalise $ cabalFiles ++ ["cabal.project", "cabal.project.local"]
```
bubba, line 397: Does `findCabalFiles` need to change into the `rootDir` here?

fendor: I dont think so, `findCabalFiles` needs to look in the sub-directory of a cradle if there is one. E.g. fir hie-plugin-api, the relevant cabal file is in the sub directory. Then it returns cabal files relative to the component directory. Afterwards we need to make the dependencies relative to the root dir, otherwise our promise that dependencies are relative to the root dir cannot be upheld.

PR #216, opened 2020-07-01
GitHub actions improvements

bubba:


.github/workflows/haskell.yml
```diff
@@ -31,14 +30,22 @@ jobs:
...
       run: cabal build
     - name: Test
       run: cabal test
+    - name: Check
+      run: cabal check
+    - name: Generate sdist
+      run: cabal sdist
+    - uses: actions/upload-artifact@v2
+      with:
+        path: dist-newstyle/sdist/hie-bios*.tar.gz
```
fendor: what haskell-ci does is to use cabal build, cabal install and cabal test on the sdist tar archive. This way it detects that a build really works after it is uploaded to hackage.

bubba: Yeah, I think its best to leave that to the travis script, since there seemed to be a lot of `sed`ing needed to get that to work. We can use this just for build artefacts of the sdist

bubba:
Ugh windows build is doing something flaky. Good to merge

fendor:
Why is it flaky? Is it a good idea to replace something working with a flaky build?

.github/workflows/haskell.yml
```diff
@@ -31,14 +30,30 @@ jobs:
         path: ~/.cabal
```
jneira, line 30: does this path works for windows? The last build is not using cache but it doesn't save it afaics

bubba: The path is working, I think it's this issue https://github.com/actions/cache/issues/144

fendor:
Merged in #267

Issue #218, opened 2020-07-02
pattern matching on symlinks use the final path of symlink ?


teto:
hi,

I have this folder with a bunch of folders and only one or 2 relevant to ghcide, e.g:
```
$ ls
_generated/
bazel-bin/ 
...
project/
```
What i would like to do is ignore all folders but "project". Hence I've written the following hie.yaml:
```
# Configuration for hie/ghcide.
# See https://github.com/mpickering/hie-bios/blob/master/README.md for the
# documentation
cradle:
    multi:
        # we default to none because of all the non-haskell folders at the root
        - path: "./"
          config: { cradle: { none:  } }
        - path: "./project"
          config:
              cradle:
                  bios:
                      program: ".hie-bios"
                      dependency-program: ".hie-bios-dependencies"
```

Now if I run `$ hie-bios debug _generated/`, I get the message `Cradle failed to load`. It looks like a real error since I see "No prefixes matched" but I would expec "./" to match ./_generated".
```
hie-bios debug _generated/
Cradle failed to load
Exit Code: ExitSuccess
Stderr: Multi Cradle: No prefixes matched
pwd: /home/teto/nova/jinko2
filepath: /nix/store/llxkyzbl2z06d5ndm2yp2cvl8fyfrxha-jinko-generated-folder
prefixes:
("./",None)
("./project",Bios {prog = ".hie-bios", depsProg = Just ".hie-bios-dependencies"})
```
Now relooking at the error, seems like it compares prefixes against the resolved filepath and because ./_generated is a symlink to /nix/store/llxkyzbl2z06d5ndm2yp2cvl8fyfrxha-jinko-generated-folder, it won't match the prefix. Is that correct ? I would expect to match against the symlink path rather its resolved path.

fendor:
Can confirm, looks like a bug to me.

I think that we are too loosely using `canonicalizePath` here: https://github.com/mpickering/hie-bios/blob/master/src/HIE/Bios/Cradle.hs#L245
Maybe a `makeAbsolute` is sufficient?

teto:
If I modify the path to
```
        - path: "./_generated"
          config: { cradle: { none:  } }
```
it works better:
```
hie-bios debug _generated/
No cradle
```
though I would prefer the message to be "Matched the None cradle" or sthg similar.

fendor:
Should be fixed in current master!

teto:
nice thank you !

PR #219, opened 2020-07-02
Dont resolve symlinks in cradle discovery

fendor:
Prefer 'makeAbsolute' over 'canonicalizePath'

Solves the problem described in #218 but I am not sure if it would break something

### DIFF
--- a/hie-bios.cabal
+++ b/hie-bios.cabal
@@ -37,6 +37,8 @@ Extra-Source-Files:     ChangeLog
                         tests/configs/obelisk.yaml
                         tests/configs/stack-config.yaml
                         tests/configs/stack-multi.yaml
+                        tests/projects/symlink-test/a/A.hs
+                        tests/projects/symlink-test/hie.yaml
                         tests/projects/multi-direct/A.hs
                         tests/projects/multi-direct/B.hs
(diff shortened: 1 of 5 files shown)


fendor:
cc @bubba would you mind to give a review here as well?

fendor:
@bubba `canonicalizePaths` resolves symlinks, which caused the issue #219. with makeAbsolute, weonly prepend the current directory and normalise it, remove indirections, etc... but do not resolve symlinks. I am not sure what could break from it and how to add a proper test-case for symlinks

bubba:
Ah, I skimmed the last crucial sentence of that issue. You can add symbolic links in git (last time I checked), so maybe just a folder like
```
- hie.yaml
- foo/ -> bar/
- bar/
```
Where hie.yaml is
```
cradle:
    multi:
        - path: "./foo"
          config: { cradle: { default:  } }
```
And then test that you can get the cradle for `/foo/A.hs`

fendor:
Added failing test-case. I decided to solve it programmatically, since I can not add this symlink to the sdist tar archive (or at least, I expect troubles and pain).

I verified that the test-case used to fail and doesnt anymore.

PR #221, opened 2020-07-03
Detect implicit cabal cradle in the absence of cabal.project

bubba:
A cabal cradle doesn't need a cabal.project file and isn't guaranteed to have one either. This improves the implicit cradle search for cabal by falling back to finding .cabal files if hie-bios couldn't find any cabal.project files.
Because it falls back, if hie-bios tries to query a multi-package cabal project with a nested cabal package, it will still find use the "larger" cradle with the upper cabal.project file. See the implicit-cabal-deep-project test project for an example of this.

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -32,7 +32,7 @@ jobs:
         key: ${{ runner.OS }}-${{ matrix.ghc }}-cabal-0
 
     - name: Run ShellCheck
-      uses: ludeeus/action-shellcheck@master
+      uses: ludeeus/action-shellcheck@2f2aa0d
       if: matrix.os == 'ubuntu-latest'
       with:
         ignore: tests
--- a/tests/BiosTests.hs
+++ b/tests/BiosTests.hs
@@ -22,6 +22,7 @@ import Data.Void
 import System.Directory
 import System.FilePath ( makeRelative, (</>) )
 import System.Info.Extra ( isWindows )
+import System.IO.Temp
 import System.Exit (ExitCode(ExitFailure))
 
 main :: IO ()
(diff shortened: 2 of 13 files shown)


mpickering:
When would the implicit stack cradle be selected with this logic?

bubba:
Always selected first, if possible. Cabal cradles are the “last resort” in the implicit lookup and this only extends the Cabal part

tests/BiosTests.hs
```diff
@@ -195,15 +206,40 @@ findCradleForModule fp expected' step = do
...
+
   step "Initialize flags"
   testLoadFile crd fp step
 
+copyToTmp :: FilePath -> (FilePath -> IO a) -> IO a
+copyToTmp srcDir f = do
+  tmpDir <- getTemporaryDirectory
+  -- canonicalize to fix https://github.com/mpickering/hie-bios/pull/221/checks?check_run_id=835539731
+  newDir <- canonicalizePath $ tmpDir </> "hie-bios-test-copy"
+  bracket_ (copyDir srcDir newDir) (removeDirectoryRecursive newDir) (f newDir)
```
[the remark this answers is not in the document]
bubba: It's not a built-in function, it's coming from a dependency 😛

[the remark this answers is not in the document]
fendor: It is in our dependency closure already: https://github.com/mpickering/hie-bios/blob/master/hie-bios.cabal#L178

tests/BiosTests.hs
```diff
@@ -195,15 +206,38 @@ findCradleForModule fp expected' step = do
   step "Finding cradle"
   findCradle a_fp `shouldReturn` expected
 
-testImplicitCradle :: FilePath -> ActionName Void -> (String -> IO ()) -> IO ()
-testImplicitCradle fp' expectedActionName step = do
-  fp <- canonicalizePath fp'
+testImplicitCradle :: FilePath -> ActionName Void -> FilePath -> (String -> IO ()) -> IO ()
+testImplicitCradle fp' expectedActionName expectedCradleRootDir step = do
```
fendor, line 210: Maybe second and third parameter should be in some correlation? E.g. we have a project dir and the filepath to load is relative to the given project dir. Otherwise we repeat ourselves quite often.

tests/BiosTests.hs
```diff
@@ -195,15 +206,38 @@ findCradleForModule fp expected' step = do
...
   actionName (cradleOptsProg crd) `shouldBe` expectedActionName
+
+  absExpectedCradleRootDir <- makeAbsolute expectedCradleRootDir
+  cradleRootDir crd `shouldBe` absExpectedCradleRootDir
+
   step "Initialize flags"
   testLoadFile crd fp step
 
+copyToTmp :: FilePath -> (FilePath -> IO a) -> IO a
+copyToTmp srcDir f =
```
fendor, line 224: Nitpick, I dont like the name (as always, this opinion can be debated)
I think I prefer something like `runTestInTmpDir` or maybe even `runTestInTemporaryDirectory`

tests/projects/implicit-cabal-deep-project/README
```diff
@@ -0,0 +1,3 @@
+Here we are testing that if we have a nested cabal package but a cabal.project
+a directory above it, hie-bios will implicitly pick up the cabal.project and
+NOT the nested cabal package
```
fendor, line 3: I like this form of documentation, could be used in different test cases as well

.github/workflows/haskell.yml
```diff
@@ -32,7 +32,7 @@ jobs:
         key: ${{ runner.OS }}-${{ matrix.ghc }}-cabal-0
 
     - name: Run ShellCheck
-      uses: ludeeus/action-shellcheck@master
+      uses: ludeeus/action-shellcheck@2f2aa0d
```
fendor, line 35: What does this do?

tests/BiosTests.hs
```diff
@@ -190,30 +202,32 @@ testLoadFile crd a_fp step = do
...
-    G.runGhc libDir $ do
-      let relFp = makeRelative (cradleRootDir crd) a_fp
-      res <- initializeFlagsWithCradleWithMessage (Just (\_ n _ _ -> step (show n))) relFp crd
-      case res of
-        CradleSuccess _ -> liftIO $ expectationFailure "Cradle loaded successfully"
-        CradleNone -> liftIO $ expectationFailure "Unexpected none-Cradle"
-        CradleFail crdlFail -> liftIO $ cradleErrorExpectation crdlFail
+  -- don't spin up a ghc session, just run the opts program manually since
+  -- we're not guaranteed to be able to get the ghc libdir if the cradle is
+  -- failing
```
fendor, line 207: That is interesting and not documented: https://github.com/mpickering/hie-bios/blob/master/src/HIE/Bios/Types.hs#L60
Or what do you mean with this?

.github/workflows/haskell.yml
bubba: Workaround for [this](https://github.com/ludeeus/action-shellcheck/issues/21)

tests/BiosTests.hs
bubba: Now that hie-bios no longer uses ghc-paths it has to call `cabal exec ghc` for the failing-cabal test case to set up the ghc session. But `cabal exec ghc` fails because it can’t resolve the constraints in the first place! So we aren’t able to get a lib dir. Prior to this it was just working by fluke, since there’s no cabal.project in this directory `cabal exec ghc` was using the hie-bios project to resolve it fine. 
We can still test loading the flags etc and it will still fail, we just need to do it without a ghc session

fendor: Maybe we should document it better that we require `runCradle` to succeed and how the results for libdir may be non-sensical if it fails?

bubba:
@fendor let's remove them when we revert the stop files commit

PR #225, opened 2020-07-08
Improve the README

### LINKED ISSUE #198: Improve Documentation
Add extensive documentation about the cabal/stack component syntax, ideally with references, more concrete examples and a nice how to write your own `hie.yaml`.

fendor:
Aims to make it easier to understand for newcomers.
It should be a nice introduction on how to write a `hie.yaml` file.
Also, it should cover most questions that arose during some hours 
of helping developers setting up their session.
If there are questions, this README should be the first reference to look for solutions.

Includes:

* Easier examples
* References to stack and cabal component syntax
* Update out-of-date information
* Add table of contents to improve discoverabilty
* Dont change existing headlines to be "backwards" compatible
    (e.g. dont make old links invalid)
* Add references in the readme for better navigation

Does not include:

* A nice glossary or a more convenient overview of common
  vocabulary.

Closes #198

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ and [`ghcide`](https://github.com/digital-asset/ghcide) set up a GHC API session
 
 Given a Haskell project that is managed by Stack, Cabal, or other package tools,
 `haskell-language-server` needs to know the full set of flags to pass to GHC in
-order to build the project. These flags might contain some compilations options
+order to build the project. These flags might contain some compilation options
 like `-O2`, but a lot of the time they are package dependencies such as
 `-package-id directory-1.3.6.0`, which also need to be built beforehand.
 `hie-bios` satisfies both these needs.
@@ -44,27 +44,27 @@ Its design is motivated by the guiding principle:
 Using this principle, it is possible
 to easily support a wide range of tools including `cabal-install`, `stack`,
 `rules_haskell`, `hadrian` and `obelisk` without major contortions.
-`hie-bios` does not depend on the `Cabal` library nor does not
+`hie-bios` does not depend on the `Cabal` library nor does it
 read any complicated build products and so on.
 
 How does a tool specify a session? A session is fully specified by a set of
 standard GHC flags. Most tools already produce this information if they support
 a `repl` command. Launching a repl is achieved by calling `ghci` with the
 right flags to specify the package database. `hie-bios` needs a way to get
-these flags and then it can set up GHC API session correctly.
+these flags and then it can set up a GHC API session correctly.
 
-Futher it means that any failure to set up the API session is the responsibility
+Further it means that any failure to set up the API session is the responsibility
 of the build tool. It is up to them to provide the correct information if they
 want the tool to work correctly.
 
 ## Explicit Configuration
 
 The user can place a `hie.yaml` file in the root of the workspace which
-describes how to setup the environment. There are several supported ways to setup the environment.
+describes how to set up the environment. There are several supported ways to setup the environment.
 
 ### Stack
 
-To explicitly state that you want to use `stack` then the basic configuration `hie.yaml` would look like:
+To explicitly state that you want to use `stack`, the basic configuration `hie.yaml` would look like:
 
 ```yaml
 cradle:
@@ -73,12 +73,12 @@ cradle:
 
 This configuration suffices if your whole project can be loaded by the command `stack repl`. As a rule of thumb, this works if the project consists of only one executable, one library and one test-suite.
 
-If your project is more complicated, you need to specify which [component](https://docs.haskellstack.org/en/stable/build_command/#components) you want to load. A component is roughly speaking a library/executable/test-suite or benchmark in `stack`. You can view the components/targets of a stack project by executing the command:
+If your project is more complicated, you need to specify which [components](https://docs.haskellstack.org/en/stable/build_command/#components) you want to load. A component is, roughly speaking, a library/executable/test-suite or benchmark in `stack`. You can view the components/targets of a stack project by executing the command:
 ``` sh
 $ stack ide targets
 ```
 
-For `hie-bios`, this would output the following:
+Since we have two test-suites, one executable and a library, for `hie-bios`, this would output the following:
 
 ``` sh
 $ stack ide targets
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -276,26 +548,29 @@ The targets are searched for in following order.
...
 which flags to provide.
 
-In this mode, an executable file called `.hie-bios` is placed in the root
-of the workspace directory. The script takes one argument, the filepath
-to the current file we want to load into the session. The script returns
-a list of GHC arguments separated by newlines which will setup the correct session.
+The program will receive the file to return options for as its first argument.
+
+The process must consult the `HIE_BIOS_OUTPUT` environment variable and write a
+list of options to this file separated by newlines. Once the process finishes
```
jneira, line 571: It is implicit and maybe obvious (with `this file`), but i would make clear that the value of `HIE_BIOS_OUTPUT` is the file path where you have to write the flags.

fendor: I agree, we should make this clear!
Also, there is another undocumented pre-condition: all targets of the component must be listed by the process.
I will extend the documentation for the bios cradle.

README.md
```diff
@@ -32,24 +56,234 @@ want the tool to work correctly.
...
+
+This way, we specify that we do not want to load any files in our test project directories.
+
+#### Internal Libraries
+
+Internal libraries are not well supported in `stack`. Since the syntax [`stack repl <internal library name>`](https://github.com/commercialhaskell/stack/issues/4564) doesn't work, `hie-bios` will generally not work with internal libraries using `stack`.
+
+### Cabal
+
+To use `cabal`, the basic explicit configuration looks similar.
```
fendor, line 158:
```suggestion
To use `cabal`, the basic explicit configuration looks similar to `stack`:
```

README.md
```diff
@@ -32,24 +56,234 @@ want the tool to work correctly.
...
+
+```yaml
+cradle:
+  cabal:
+```
+
+The implication of this configuration is a bit different, though. Given a source file to load, we will use `cabal repl <filename>` to find the component of the given filepath.
+
+Due to an unfortunate [bug](https://github.com/haskell/cabal/issues/6622), this fails on some files with `cabal` versions older than `3.4`.
```
fendor, line 167:
```suggestion
This configuration should work in (almost) every standard project setup, since `cabal` finds the component associated to a given source file.
However, due to an unfortunate [bug](https://github.com/haskell/cabal/issues/6622), this fails on some files if the `cabal` version is older than `3.4`. 
```

README.md
```diff
@@ -32,24 +56,234 @@ want the tool to work correctly.
...
+
+```yaml
+cradle:
+  cabal:
+    component: "lib:hie-bios"
 ```
 
-While the component is optional, this is recommended to make sure the correct
-component is loaded.
+The component syntax `"lib:hie-bios"` refers to the library of the package `hie-bios`. For a complete reference of the component syntax, we refer to the [documentation](https://cabal.readthedocs.io/en/stable/nix-local-build.html?highlight=component#cabal-v2-build).
```
fendor, line 188: This should make it why the given configuration isnt enough usually

README.md
```diff
@@ -32,24 +56,234 @@ want the tool to work correctly.
...
+              component: "exe:hie-bios"
+            - path: "./tests/BiosTests.hs"
+              component: "test:hie-bios"
+            - path: "./tests/ParserTests.hs"
+              component: "test:parser-tests"
+```
+
+This way, we specify that we do not want to load any files in our test project directories.
+
+### Bios
```
fendor, line 285: Section needs to be improved

README.md
```diff
@@ -63,26 +297,50 @@ list of options to this file separated by newlines. Once the process finishes
...
-is most useful when combined with the multi-cradle which is specified in the next section.
+is most useful when combined with the [multi-cradle](#multi-cradle) which is specified in the next section.
 
 ```yaml
-cradle: {none: }
+cradle:
+  none:
 ```
 
 ## Multi-Cradle
```
fendor, line 346: Maybe a "how to debug" section is useful here, too.

README.md
```diff
@@ -32,24 +56,234 @@ want the tool to work correctly.
 ## Explicit Configuration
 
 The user can place a `hie.yaml` file in the root of the workspace which
-describes how to setup the environment. For example, to explicitly state
-that you want to use `stack` then the configuration file would look like:
+describes how to setup the environment. There are several supported ways to setup the environment.
+
+### Stack
+
+To explicitly state that you want to use `stack` then the basic configuration file would look like:
```
fendor, line 63:
```suggestion
To explicitly state that you want to use `stack` then the basic configuration `hie.yaml` would look like:
```

README.md
```diff
@@ -32,24 +56,234 @@ want the tool to work correctly.
...
 
 ```yaml
-cradle: {stack: {component: "haskell-ide-engine:lib" }}
+cradle:
+  stack:
+    component: "<component name>"
+```
+
+where `<component name>` is the name of component/target you want to be able to load.
+While the component is optional, this is recommended to make sure the correct component is loaded.
```
fendor, line 98: Add a motivating example, why do we usually need the next section

README.md
```diff
@@ -1,12 +1,36 @@
...
-[`hie`](https://github.com/haskell/haskell-ide-engine) and
-[`ghcide`](https://github.com/digital-asset/ghcide) sets up a GHC API session.
+[`haskell-language-server`](https://github.com/haskell/haskell-language-server)
+and [`ghcide`](https://github.com/digital-asset/ghcide) set up a GHC API session.
 
 Given a Haskell project that is managed by Stack, Cabal, or other package tools,
-`hie` needs to know the full set of flags to pass to GHC in order to build the
-project. `hie-bios` satisfies this need.
+`haskell-language-server` needs to know the full set of flags to pass to GHC in
+order to build the project. `hie-bios` satisfies this need.
```
bubba, line 33: Just to emphasise that the build tools need to do more than just return the flags, but also properly set up the environment by building deps etc
```suggestion
`haskell-language-server` needs to know the full set of flags to pass to GHC in
order to build the project. These flags might contain some compilations options
like `-O2`, but a lot of the time they are package dependencies such as
`-package-id directory-1.3.6.0`, which also need to be built beforehand.
`hie-bios` satisfies both these needs.
```

jneira:
Maybe a little bit too late but, what do you think about mentioning `implicit-hie` as a helper to create a explicit `hie.yaml` for stack or cabal?

fendor:
Oh yeah, shoot, completely forgot about that (I actually was planning that). However, I think it is fine here, we should mention it more prominently on HLS/ghcide.

Issue #44, opened 2019-09-16
Add extra explicit arguments to call cradles


ndmitchell:
As requested in https://github.com/digital-asset/ghcide/issues/62#issuecomment-531934756, it would be nice to write:

```
cradle:
    stack:
    arguments:
        - -Wunused-binds
        - -Wunused-imports
```

Seems like it should apply to all cradles.

ndmitchell:
CC @fendor @jacg

mpickering:
I think that you should add these options to your `stack.yaml`?

jacg:
It works, but it the warnings I want in my IDE aren't necessarily the same options I want during compilation.

fendor:
I think, we have to be careful to not reinvent the wheel with the specification of `hie.yaml`.
If we continue to rapidly adding features, we might soon have an unmaintainable behemoth.

However, I think it is worth implementing custom arguments in one way or another. For example, in the case of a stack cradle, if projects support more than one snapshot, e.g. has multiple `stack-*.yaml` files, hie-bios automatically uses `stack.yaml`. Let's assume this `stack.yaml` uses ghc 8.6.5. However, sometimes I would rather want to use `stack-8.6.4.yaml`, since my version of `hie`/`ghcide` is compiled with ghc 8.6.4.
Currently, to work around this, we can copy `stack-8.6.4.yaml` to `stack.yaml`. However, now we have to careful to not commit the changes to `stack.yaml` and revert these changes if we want to switch the branch, which is not optimal.
Thus, in my opinion, it would be useful if we can specify the `stack.yaml` to use in `hie.yaml`. Consequentially, custom arguments for the cradle build-tools might make sense.

jacg:
I'm certainly very uncomfortable with having to edit a project's version-controlled files, just to get my IDE to behave the way I want it to, and then to have to juggle these edits in stashes whenever I want to switch branch or rebase.

fendor:
I want to try tackle this feature. 
We will have to discuss a potential API.

vilem-ff:
+1 for allowing different GHC options than those listed in the `stack.yaml`. At the moment I have to make sure I don't accidentally commit my `stack.yaml` and/or don't forget to commit those changes from `stack.yaml` that I did actually mean to commit.

fendor:
I have been thinking about this feature request.
[...]

PR #230, opened 2020-07-31
Allow specifying a stack.yaml for stack configurations

WorldSEnder:
In response to #44 this PR adds an argument to `Stack/StackMulti` configurations for providing an explicit `stack.yaml` file to use. This is hard to do with generic additional arguments as mentioned in #44 since `stackCradleDependencies` has to know about it, so I decided to give it special syntax. Example hie.yaml (which I'd use for [agda/agda](https://github.com/agda/agda))

```
cradle:
  stack:
    - path: "./src/full"
      component: Agda:lib
      stackYaml: stack-8.8.3.yaml
    - path: "./src/main"
      component: Agda:exe:agda
      stackYaml: stack-8.8.3.yaml
```

A few questions remain, of course, hence this is a draft.
- Naming for `stackYaml` could simply be `yaml`
- Documentation and test cases are currently missing
- A small semantic change is that multistack/multicabal configurations are now allowed to only have a `path` key, defaulting the rest, not requiring a `component`. Imo, this is better behaviour anyway.

### DIFF
--- a/src/HIE/Bios/Config.hs
+++ b/src/HIE/Bios/Config.hs
@@ -2,12 +2,18 @@
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE TupleSections #-}
 {-# LANGUAGE DeriveFunctor #-}
+{-# LANGUAGE PatternSynonyms #-}
+{-# LANGUAGE LambdaCase #-}
 -- | Logic and datatypes for parsing @hie.yaml@ files.
 module HIE.Bios.Config(
     readConfig,
     Config(..),
     CradleConfig(..),
+    CabalType(..),
+    StackType(..),
     CradleType(..),
+    pattern Cabal,
+    pattern Stack,
     Callable(..)
     ) where
 
@@ -43,11 +49,25 @@ data CradleConfig a =
 data Callable = Program FilePath | Command String
     deriving (Show, Eq)
 
+data CabalType
+    = CabalType { cabalComponent :: Maybe String }
+    deriving (Eq)
+
+instance Show CabalType where
+  show = show . Cabal_
+
+data StackType
+    = StackType { stackComponent :: Maybe String , stackYaml :: Maybe String }
+    deriving (Eq)
+
+instance Show StackType where
+  show = show . Stack_
+
 data CradleType a
-    = Cabal { component :: Maybe String }
-    | CabalMulti [ (FilePath, String) ]
-    | Stack { component :: Maybe String }
-    | StackMulti [ (FilePath, String) ]
+    = Cabal_ { cabalType :: !CabalType }
+    | CabalMulti [ (FilePath, CabalType) ]
+    | Stack_ { stackType :: !StackType }
+    | StackMulti [ (FilePath, StackType) ]
 --  Bazel and Obelisk used to be supported but bit-rotted and no users have complained.
 --  They can be added back if a user
 --    | Bazel
@@ -66,6 +86,14 @@ data CradleType a
     | Other { otherConfig :: a, originalYamlValue :: Value }
     deriving (Eq, Functor)
 
+pattern Cabal :: Maybe String -> CradleType a
+pattern Cabal cm = Cabal_ (CabalType cm)
+
+pattern Stack :: Maybe String -> Maybe String -> CradleType a
+pattern Stack cm yml = Stack_ (StackType cm yml)
+
+{-# COMPLETE Cabal, CabalMulti, Stack, StackMulti, Bios, Direct, None, Multi, Other :: CradleType #-}
+
 instance FromJSON a => FromJSON (CradleType a) where
     parseJSON (Object o) = parseCradleType o
     parseJSON _ = fail "Not a known cradle type. Possible are: cabal, stack, bios, direct, default, none, multi"
@@ -73,7 +101,7 @@ instance FromJSON a => FromJSON (CradleType a) where
 instance Show (CradleType a) where
     show (Cabal comp) = "Cabal {component = " ++ show comp ++ "}"
     show (CabalMulti a) = "CabalMulti " ++ show a
-    show (Stack comp) = "Stack {component = " ++ show comp ++ "}"
+    show (Stack comp yaml) = "Stack {component = " ++ show comp ++ ", stackYaml = " ++ show yaml ++ "}"
     show (StackMulti a) = "StackMulti " ++ show a
     show Bios { call, depsCall } = "Bios {call = " ++ show call ++ ", depsCall = " ++ show depsCall ++ "}"
     show (Direct args) = "Direct {arguments = " ++ show args ++ "}"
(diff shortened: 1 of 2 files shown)


jneira:
Unfortunately `stack` shoud be a list *or* key-values.
My ideal shape would be
```yaml
cradle:
  stack:
    config: stack-8.8.3.yaml
    components:
      - path: "./src/full"
        target: Agda:lib
      - path: "./src/main"
        target: Agda:exe:agda
```
But i would be out of scope of this pr...

WorldSEnder:
Paths are supposed to be relative to the `hie.yaml` file, which also is the working directory of the `stack` command. I favor @jneira's config shape, and I suppose the following would work during parsing:

- Look at the Value under `stack:`
- If it's a list, do what has already been implemented in this PR.
- If it's a an object with a `components:` key, parse those as a list of configurations, parse the rest of the object as a single configuration, then monoidally combine these configurations, where the more specific options overwrite the options in the parent one.
- If there is no `components:`, parse it as a single configuration.

WorldSEnder:
> Also, are these paths expected to be relative to the cradle root directory?
> What would happen in the face of nested multi-cradles, such as:
> 
> ```yaml
> cradle:
>   multi:
>     - path: "appA"
>       config:
>         cradle:
>           stack:
>             - path: ./src
>               component: appA:lib
>               stackYaml: stack-8.8.3.yaml
>     - path: "appB"
>       config:
>         cradle:
>           stack:
>             - path: ./src
>               component: appB:lib
>               stackYaml: stack-8.8.3.yaml
> ```
> 
> I guess in this case, the `stackYaml` needs to be relative to the cradleRootDir, otherwise we can not express the case that `appA` and `appB` have the same parent `stack.yaml` file.

I added a test case with this config, where I needed to say `-path: appA/src` instead of `./src`. Not sure if this is intended behaviour, have a look once I pushed. So it seems that your guess is already the behaviour implemented. To be concrete, the `stackYaml` is relative to the working directory, which is, as far as I can tell, the `cradleRootDir`.

```yaml
cradle:
  multi:
    - path: "appA"
      config:
        cradle:
          stack:
            - path: appA/src
              component: appA:lib
              stackYaml: stack-8.8.3.yaml
    - path: "appB"
      config:
        cradle:
          stack:
            - path: appB/src
              component: appB:lib
              stackYaml: stack-8.8.3.yaml
```

WorldSEnder:
I'm a bit lost as to why the 8.10.1 test cases are failing, after all. I use ghc 8.10.1 locally, too. Apart from [this issue](https://github.com/commercialhaskell/stack/issues/5134) the test appear just fine. Secondly, the Travis CI build seems to not find files, that the other builds do find, what is going on there?

README.md
```diff
@@ -126,6 +136,24 @@ Here you can see two important features:
...
+    components:
+    - path: "./src"
+      component: "hie-bios:lib"
+    - path: "./exe"
+      component: "hie-bios:exe:hie-bios"
+    - path: "./tests/BiosTests.hs"
+      component: "hie-bios:test:hie-bios"
+    - path: "./tests/ParserTests.hs"
+      component: "hie-bios:test:parser-tests"
+```
```
fendor, line 155: It is important to note that all `stackYaml` files *must* point to the same GHC version.
Otherwise, confusions such as https://github.com/haskell/haskell-language-server/issues/268 are bound to happen

WorldSEnder: That's a problem that needs to be fixed upstream in the language server. It would certainly be possible to find out which ghc version is in use for the stack config (`stack [args] ghc --version`), use multiple `.stack-work` directories and load different ghcide plugins (if I understand the architecture correctly). I will add a note though with the rest of the review, for now.

WorldSEnder:
Ping @fendor

src/HIE/Bios/Config.hs
```diff
@@ -43,11 +65,45 @@ data CradleConfig a =
...
+instance Show StackType where
+  show = show . Stack
+
 data CradleType a
-    = Cabal { component :: Maybe String }
-    | CabalMulti [ (FilePath, String) ]
-    | Stack { component :: Maybe String }
-    | StackMulti [ (FilePath, String) ]
+    = Cabal { cabalType :: !CabalType }
+    | CabalMulti { defaultCabal :: !CabalType, subCabalComponents :: [ (FilePath, CabalType) ] }
```
fendor, line 104: This also allows the config file:
``` yaml
cradle:
  cabal:
    components:
      - path: "./src"
        component: "lib:hie-bios"
      - path: "./"
        component: "lib:hie-bios"
```

While I dont expect any harm of it (and maybe we want soonish a similar field for cabal) but I just wanted to document it.

src/HIE/Bios/Config.hs
```diff
@@ -96,38 +152,61 @@ parseCradleType o
...
-parseStackOrCabal singleConstructor _ (Object x)
-  | Map.size x == 1, Just (String stackComponent) <- Map.lookup "component" x
-  = return $ singleConstructor $ Just $ T.unpack stackComponent
-  | Map.null x
-  = return $ singleConstructor Nothing
-  | otherwise
-  = fail "Not a valid Configuration type, following keys are allowed: component"
-parseStackOrCabal _ multiConstructor (Array x) = do
-  let parseOne e
+parseSingleOrMultiple single multiple parse = doParse where
```
fendor, line 162: Interestingly, this allows this configuration: 
``` yaml
cradle:
  cabal:
    components:
      - path: "./src"
        component: "lib:hie-bios"
      - path: "./"
        component: "lib:hie-bios"
    components:
      - path: "./src"
        component: "exe:hie-bios"
      - path: "./"
        component: "exe:hie-bios"
``` 
with this result:
``` haskell 
Config
  { cradle =
      CradleConfig
        { cradleDependencies = []
        , cradleType =
            CabalMulti
              { defaultCabal = Cabal { component = Nothing }
              , subCabalComponents =
                  [ ( "./src" , Cabal { component = Just "exe:hie-bios" } )
                  , ( "./" , Cabal { component = Just "exe:hie-bios" } )
                  ]
              }
        }
  }
```

bubba:
This makes a lot of sense to have, what is the `stackYaml` path relative to? To me I would intuitively expect it to be relative to the `hie.yaml` file, not the component/cradle root

src/HIE/Bios/Config.hs
WorldSEnder: That's a good catch, didn't think about multiple `components` entries. I'll fix this before the PR is merged.

WorldSEnder:
@bubba Maybe I'm missing something but I was basing the consideration on the implemented reality in [`multiCradle`](https://github.com/mpickering/hie-bios/blob/64a12edd7ab544dbb1e0685c53fe1e6c4fa4ec86/src/HIE/Bios/Cradle.hs#L242). The `cur_dir` is taken as the base path to append the component sub-paths `cur_dir </> p` off of. As I read it, this is relative to the cradle root, then. I would also agree with the suggestion to unambiguously make all paths inside `hie.yaml` relative to the file itself.

Additionally, it seems that the main entry point `loadCradleWithOpts` simply takes the directory containing `hie.yaml` as the cradle root directory, so it seems that relative to cradle root and relative to `hie.yaml` often means the same thing.

src/HIE/Bios/Config.hs
WorldSEnder: Thinking about this a little bit more, isn't a duplicate key (`components`) an error according to the [yaml spec](https://yaml.org/spec/1.1/#id861060)?

> Mapping nodes are somewhat tricky because their keys are unordered and must be unique.

In that case, I suppose [this PR](https://github.com/snoyberg/yaml/pull/146/files) is relevant and https://github.com/mpickering/hie-bios/pull/230/files#diff-0478bef4574ee638b0c136cccc31479aR288 should use [`decodeFileWithWarnings`](https://www.stackage.org/haddock/lts-16.10/yaml-0.11.4.0/Data-Yaml.html#v:decodeFileWithWarnings)

bubba:
@WorldSEnder 
>  I would also agree with the suggestion to unambiguously make all paths inside hie.yaml relative to the file itself.
This is probably worthy of a separate issue and discussion, but for now in this PR probably best then to just stick with making it relative to the `cur_dir` for consistency

src/HIE/Bios/Config.hs
WorldSEnder: I added a test case that's supposed to fail on the config with duplicate `components` key.

PR #231, opened 2020-08-07
Add optional ghc-path field in bios cradles

### LINKED ISSUE #215: How should bios cradles run ghc?
This a further piece of work from #207:

Cabal and stack cradles can run GHC to get the libdir from `cabal exec ghc -- --print-libdir` and `stack exec ghc -- --print-libdir` respectively.
Direct cradles just use the `ghc` on the path.
But what about bios programs? What ghc are they using exactly? There should be a way for the bios program to report this, either via the program or in a hie.yaml file

pepeiborra:
Fixes #215 to a simpler extent, in that the `with-ghc` field is of type `String` rather than `IO String`.


src/HIE/Bios/Cradle.hs
```diff
@@ -334,14 +334,14 @@ directCradle wdir args =
...
-biosCradle wdir biosCall biosDepsCall =
+biosCradle :: FilePath -> Callable -> Maybe Callable -> Maybe FilePath -> Cradle a
+biosCradle wdir biosCall biosDepsCall mbGhc =
   Cradle
     { cradleRootDir    = wdir
     , cradleOptsProg   = CradleAction
         { actionName = Types.Bios
         , runCradle = biosAction wdir biosCall biosDepsCall
-        , runGhcCmd = runGhcCmdOnPath wdir
+        , runGhcCmd = \args -> readProcessWithCwd wdir (fromMaybe "ghc" mbGhc) args ""
```
fendor, line 344: Would it make sense to canonicalize the path if it is relative?

README.md
```diff
@@ -355,6 +355,14 @@ cradle:
     shell: "<build-tool flags $HIE_BIOS_ARG>"
 ```
 
+Additionally, you may specify the path to ghc. Otherwise, the one in the PATH will be used:
+
+```yaml
+cradle:
+  bios:
+    program: "<program>"
+    with-ghc: :<ghc>"
```
fendor, line 364:
```suggestion
    with-ghc: "<ghc>"
```

src/HIE/Bios/Cradle.hs
pepeiborra: Good question, I'm not sure. The path is only ever used in this call to `readProcess`, so I think it probably doesn't matter much.

pepeiborra:
Thanks for the quick review! 

Once I fix the test failures, I'm very keen on a minor release incorporating this change as it will allow us to start using ghcide in some parts of our Haskelll codebase at work. Thanks again!

fendor:
The travis tests fail because you need to add the test files to the `Extra-Source-Files` section.

PR #233, opened 2020-08-12
Add tests for bios dependencies

pepeiborra:
I ran into issues trying to define component dependencies in my Bios cradles, and wrote a pair of failing tests to demonstrate. 
Please check that I'm not missing anything, and let me know if there is an obvious fix.

PR #235, opened 2020-08-12
HIE_BIOS_ARG not passed to the dependencies program

### LINKED ISSUE #234: Bios dependencies program is not passed at target as argument
Actually, there is another issue for me. The documentation says that `dependency-program` is called with no arguments, but that is inconsistent with `program` and doesn't really work for us. Is that by accident or by design?

_Originally posted by @pepeiborra in https://github.com/mpickering/hie-bios/pull/233#issuecomment-672847740_

pepeiborra:
Follow up to #233 in order to fix #234

### DIFF
--- a/README.md
+++ b/README.md
@@ -572,8 +572,8 @@ cradle:
 ```yaml
 cradle:
   bios:
-    shell: build-tool flags $HIE_BIOS_ARG $HIE_BIOS_OUTPUT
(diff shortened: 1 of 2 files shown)


fendor:
I have no idea why the tests fail... maybe github does not understand cmd? O

fendor:
The tests are broken on windows in general... So, unrelated to your changes.

pepeiborra:
This is ready to merge (and release please)

fendor:
Release will be later this day

PR #243, opened 2020-08-31
Add explicit type for stack.yaml location

fendor:
Also, if no stack.yaml was given, don't default to "stack.yaml"

This also fixes a regression. Assume the following setup:
``` sh
> cat File.hs
module File where

foo :: Int
foo = 2
> cat hie.yaml
cradle:
  stack:
> hie-bios check File.hs
```

This used to work, as we simply invoked `stack repl` to find the project arguments

### DIFF
--- a/src/HIE/Bios/Cradle.hs
+++ b/src/HIE/Bios/Cradle.hs
@@ -31,7 +31,7 @@ import System.Exit
 import HIE.Bios.Types hiding (ActionName(..))
 import qualified HIE.Bios.Types as Types
 import HIE.Bios.Config
-import HIE.Bios.Environment (getCacheDir)
+import HIE.Bios.Environment (getCacheDir, prependIfRelative)
 import System.Directory hiding (findFile)
 import Control.Monad.Trans.Maybe
 import System.FilePath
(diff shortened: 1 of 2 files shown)


fendor:
cc @WorldSEnder might be interesting for you. Also, this is the only way I can ask for a review from you :)

WorldSEnder:
That's a good catch, I didn't think about standalone stack projects. I suppose there should be a testcase to document the regression.

I have a question about the usage of `prependIfRelative`. I was under the expression that [</>](https://www.stackage.org/haddock/lts-16.12/filepath-1.4.2.1/System-FilePath-Posix.html#v:-60--47--62-) already handles this, by returning the second argument if it starts with a drive letter or directory separator. Is there some windows detail I'm missing? In any case, I think it would be worth to document this with a test case (in a different PR).

fendor:
>  I was under the expression that </> already handles this 

You seem to be right! Then this is an oversight, and this function should not exist at all. We should remove it. Thank you for mentioning this! I was not aware that this should work.

WorldSEnder:
There are some subtle differences, if I read the documentation correctly (I don't have a windows system to test right now). For example, `isRelative "c:test" == True` but `"D:\\foo" </> "C:bar" == "C:bar"`.

But I don't see an example where `isAbsolute f` and `(wdir </> f) /= f`, so I do believe `prependIfRelative` is superfluous.

PR #249, opened 2020-09-13
Require aeson 1.4.5

ndmitchell:
We rely on JSONPath, which only got exposed in 1.4.5.


fendor (approved):
Thank you for catching it! I guess this will be another hackage revision

PR #258, opened 2020-10-29
Modify unreachable cabal website links

July541:
Hello,
This pr is about the modification of cabal docs as they have changed their links from `stable` to `latest`, the previous is not available, I made it correct.


jneira:
Hi, thanks for th pr. However, i just tried the url https://docs.haskellstack.org/en/stable/build_command/#components and it works for me

July541:
I'm sorry I process them incorrectly, may I pull a new request to fix that or any other appropriate solutions you prefer?

PR #259, opened 2020-10-29
Modify unreachable cabal website links

July541:
Hello,
This pr is about the modification of cabal docs as they have changed their links from stable to latest, the previous is not available, I made it correct.


jneira (approved):
@July541 many thanks for the correction
(btw you could had changed the previous pr with a force push and renaming it in github but open a new one it is fine too)

jneira:
Just for being complete:
* urls were changed following the cabal issue: https://github.com/haskell/cabal/issues/6966
* afaiu the disabling of stable urls is temporary and they will be restored
* we should return to stable urls as soon they are available

July541:
> @July541 many thanks for the correction
> (btw you could had changed the previous pr with a force push and renaming it in github but open a new one it is fine too)

Thank you for your advice to a rookie!

fendor (approved):
Thank you!

PR #264, opened 2020-11-11
Set builddir for cabal

### LINKED ISSUE #195: Cabal cradle causes me to constantly need to run cabal configure
I'm using ghcide 0.2 with Cabal cradles. I now seem to constantly having to do a Cabal configure everytime I do `cabal build`, because opening new files in causes ghcide to run Cabal configure through hie-bios. The configuration phase takes a substantial amount of time to run (we're talking minutes), so I now seem to spend most of my time watching Cabal configure itself. I imagine this is because hie-bios uses `--with-compiler` which requires a full recnofiguration. Could this be done in such a way that it doesn't disturb what I do in my working directory?

wz1000:
Should fix #195

### DIFF
--- a/src/HIE/Bios/Cradle.hs
+++ b/src/HIE/Bios/Cradle.hs
@@ -26,6 +26,7 @@ module HIE.Bios.Cradle (
 import Control.Exception (handleJust)
 import qualified Data.Yaml as Yaml
 import Data.Void
+import Data.Char (isSpace)
 import System.Process
 import System.Exit
 import HIE.Bios.Types hiding (ActionName(..))
@@ -490,7 +491,10 @@ withCabalWrapperTool (mbGhc, ghcArgs) wdir k = do
 cabalAction :: FilePath -> Maybe String -> LoggingFunction -> FilePath -> IO (CradleLoadResult ComponentOptions)
 cabalAction work_dir mc l fp = do
   withCabalWrapperTool ("ghc", []) work_dir $ \wrapper_fp -> do
-    let cab_args = ["v2-repl", "--with-compiler", wrapper_fp, fromMaybe (fixTargetPath fp) mc]
+    abs_work_dir <- makeAbsolute work_dir
+    let dirHash = show (fingerprintString abs_work_dir)
+    buildDir <- getCacheDir ("dist-"<>filter (not . isSpace) (takeBaseName abs_work_dir)<>"-"<>dirHash)
+    let cab_args = ["--builddir="<>buildDir,"v2-repl", "--with-compiler", wrapper_fp, fromMaybe (fixTargetPath fp) mc]
     (ex, output, stde, args) <-
       readProcessWithOutputFile l work_dir (proc "cabal" cab_args)
     case processCabalWrapperArgs args of
--- a/src/HIE/Bios/Cradle.hs
+++ b/src/HIE/Bios/Cradle.hs
@@ -401,15 +401,13 @@ cabalCradle wdir mc =
         { actionName = Types.Cabal
         , runCradle = cabalAction wdir mc
         , runGhcCmd = \args -> do
+            buildDir <- cabalBuildDir wdir
             -- Workaround for a cabal-install bug on 3.0.0.0:
             -- ./dist-newstyle/tmp/environment.-24811: createDirectory: does not exist (No such file or directory)
-            -- (It's ok to pass 'dist-newstyle' here, as it can only be changed
-            -- with the --builddir flag and not cabal.project, which we aren't
-            -- using in our call to v2-exec)
-            createDirectoryIfMissing True (wdir </> "dist-newstyle" </> "tmp")
+            createDirectoryIfMissing True (buildDir </> "tmp")
             -- Need to pass -v0 otherwise we get "resolving dependencies..."
             readProcessWithCwd
-              wdir "cabal" (["v2-exec", "ghc", "-v0", "--"] ++ args) ""
+              wdir "cabal" (["--builddir="<>buildDir,"v2-exec", "ghc", "-v0", "--"] ++ args) ""
         }
     }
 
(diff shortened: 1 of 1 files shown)


fendor:
cc @jneira, @bubba , @mpickering

fendor:
I think this is a good compromise, it fixes #195, with no costs for the future.

wz1000:
What needs to change here?

```haskell
  Cradle
    { cradleRootDir    = wdir
        { actionName = Types.Cabal
        , runCradle = cabalAction wdir mc
        , runGhcCmd = \args -> do
            -- Workaround for a cabal-install bug on 3.0.0.0:
            -- ./dist-newstyle/tmp/environment.-24811: createDirectory: does not exist (No such file or directory)
            -- (It's ok to pass 'dist-newstyle' here, as it can only be changed
            -- with the --builddir flag and not cabal.project, which we aren't
            -- using in our call to v2-exec)
            createDirectoryIfMissing True (wdir </> "dist-newstyle" </> "tmp")
            -- Need to pass -v0 otherwise we get "resolving dependencies..."
            readProcessWithCwd
              wdir "cabal" (["v2-exec", "ghc", "-v0", "--"] ++ args) ""
        }
    }
```

jneira:
Not sure if always use another `builddir` without user explicit consent could drive to unexpected results (for the user) when they change config files (`cabal.project`, `.cabal`) f.e.
The builds could be out of sync and the user will not know why. Build products will be duplicated and they will not know it either.

Now hie-bios is using `with-compiler` and afaik it makes part of the cache stale and trigger the rebuild. Not sure if it still reuses something (the deps plan?).
But when `show-build-info` lands the main build could be reused entirely, so we will have to remove this to no duplicate it without no gain. So it will have costs in the future imo. 

I would prefer to, at least, use a environment variable (`HIE_BIOS_BUILD_DIR`) and use it for both stack and cabal (for being symmetric) only if the users have set it explicitly, in the location they choose.

bubba:
Has anyone combed through ghcide/hls/the rest of his-bios to see if we've hardcoded dist-newstyle anywhere?

wz1000:
> The builds could be out of sync and the user will not know why

I don't think so - nix style builds are supposed to be deterministic given a `cabal.project(.local)` and a index state.

> Build products will be duplicated and they will not know it either.

`ghcide` already duplicates build products in `~/.cache/ghcide`. This is a very reasonable thing to do, and the only way to ensure the IDE doesn't mess with the usual build configuration. There have been lots of bug reports over the years about running `cabal`/`stack` concurrently with the IDE messing things up. This should fix that.

wz1000:
> Has anyone combed through ghcide/hls/the rest of his-bios to see if we've hardcoded dist-newstyle anywhere?

I took a quick look and this was the only other concerning occurrence I found.

wz1000:
>  But when show-build-info lands the main build could be reused entirely, so we will have to remove this to no duplicate it without no gain. So it will have costs in the future imo.

I think we want to keep this even with `show-build-info` for the aforementioned bug reports about running the build concurrently with the IDE.

fendor:
Using an environment variable seems sensible to me, and I agree it makes sense to have compiler artifacts for the IDE separated from the rest. IMO, `cabal` should handle it correctly, sooner or later (e.g. create a ide directory within `dist-newstyle`).

I am currently rather opposed to adding fields to `hie.yaml` unless they really make sense. Adding fields for working around bugs, feels not like well design and I fear it will render the `hie.yaml` specification unwieldy and hard to change in the future.

Neither HLS nor ghcide will use a hard-coded `dist-newstyle`, otherwise it should be changed ASAP anyways, as it breaks the abstraction provided by `hie-bios`.

jneira:
> I don't think so - nix style builds are supposed to be deterministic given a cabal.project(.local) and a index state.

I worry about the "are supposed to be" part. But also by the fact user commands like `cabal clean` will no affect the hie-bios build, so users will have to look for the cache directory to delete it manually (if they know what have to do). I see you has included the absolute path in that directory to help found it.

I cant think of more concrete issues now but i guess that they could arise. I have to admit it doesn't seem like a convincing argument

> here have been lots of bug reports over the years about running cabal/stack concurrently with the IDE messing things up. This should fix that.

But that bug is not deterministic, right? I did not hit it frequently myself, so we can offer the env var as a temporary workaround for a bug that should be fixed.

> I think we want to keep this even with show-build-info for the aforementioned bug reports about running the build concurrently with the IDE.

I dont know much about s-b-i internals but i expect it will query lazily all possible info and only trigger update commands if needed, so it is supposed to trigger less that bug. In that case we will want to give it all possible info previously generated by the user so the default will have to be to share the same builddir, imo.

> I am currently rather opposed to adding fields to hie.yaml unless they really make sense. Adding fields for working around bugs, feels bad design and I fear it will render the hie.yaml specification unwieldy and hard to change in the future.

Fair enough, the configuration (and its interactions) can grow exponentially if we dont consider carefully additions. I would add the `project-file` anyways, to make the api symmetric between stack and cabal and wait (or submit a pr!) to make cabal honour `builddir` in the project file. To have a way to keep different settings for the ide seems to be useful.

fendor:
> I would add the project-file anyways, to make the api symmetric between stack and cabal

Yeah, that seems sensible to me, too.

> by the fact user commands like cabal clean will no affect the hie-bios build

True, but stuff like that already happens as the .hie and .hi files are lcoated in `~/.cache/ghcide/...` already.

jneira:
> True, but stuff like that already happens as the .hie and .hi files are lcoated in ~/.cache/ghcide/... already.

Well, that files are not generated by cabal directly but they belongs to ghc, no? `.hie` ones are not even generated by default in my cabal builds. Handle ghc is a ghcide business so makes sense that it put them in a cache, independent of the build tools (and moreover if those files are not being generated by default or for some ghc versions).

I think the problematic ones would be the generated directly by cabal (`cache` dir, pakagedb's, `setup-config`, etc)
I remember i had to delete the ghcide cache file only on time some time ago but i do `cabal clean`'s daily.

Maybe i am being too cautious without real reasons so go ahead if you think it will be reliable enough. 😃

fendor:
On Linux, I absolutely never have to execute `cabal clean`. The only issue I encountered was that my cabal store got corrupted by the internal libraries thingy of cabal-helper. 
Imo, this is the only thing blocking this PR right now, how often does that happen and can the IDE get stuck because of it.

fendor (approved):

fendor:
cc @ndmitchell

pepeiborra:
I am supportive of this change

jneira (approved):
I am fine with this then. If we hit real issues thereafter we can make it opt-in or at least opt-out.

wz1000:
This should be ready now, please review again.

wz1000:
Hmm https://github.com/commercialhaskell/stack/issues/2954/

jneira:
Adding suppor for stack has another caveat: hie-bios is gonna ignore the `work-dir` set in the `stack.yaml` users can give to hie-bios via `stack-yaml` field in `hie.yaml`.

We are using it as a workaround for some issues and this would fix it but i would not use it if user is using that field (`stack-yaml`).

wz1000:
I have reverted the stack change, someone who uses it can pick it back up if the want.

fendor (approved):
LGTM

michaelpj:
I'd love to have a Hackage release with this in.

fendor:
This is blocked on the current behaviour of ghcide/hls. With this PR, `Paths_*` modules can't be properly found and you are greeted with a lot of "No cradle found for `Paths_*` module" and "No cradle found for..." messages.

I am pretty sure, the issue is how we cache results in https://github.com/haskell/haskell-language-server/blob/master/ghcide/session-loader/Development/IDE/Session.hs#L122 but I got side-tracked by the holidays and so on.
Essentially, Ghcide asks for each module where its `hie.yaml` is and then but for `Paths_*` in some `~/.cache/.../Paths_*` it can't find the appropriate `hie.yaml`, therefore the message.

A fix could be to maintain an additional cache for all FilePaths to their respective ComponentOptions directly. 

I plan to get back to it this weekend.

PR #268, opened 2020-11-18
Fix ci 2020 11 18

fendor:


ndmitchell (approved):
Would be nice if we could turn Appveyor off...

fendor:
That's something @mpickering has to do.

jneira:
> That's something @mpickering has to do.

removing the config file does not disable it? ugh

fendor:
@jneira I hoped so as well, but apparently not.

Issue #270, opened 2020-12-13
Proposal: faster Bios protocol


pepeiborra:
Currently the Bios protocol can involve two process calls: one asking for the flags/targets and another asking for the deps. This split means that the Bios program may need to implement some stateful caching in order to respond efficiently to the second request. 

The proposal is to merge both calls into a single one using two env vars: `HIE_BIOS_FLAGS` and `HIE_BIOS_DEPS`. I would add a third output value, the path to ghc, in an env var `HIE_BIOS_GHC_PATH`.

wz1000:
`HIE_BIOS_GHC_PATH` will also unblock #265

pepeiborra:
Is anyone working on `HIE_BIOS_GHC_PATH`? It's the only piece missing and I might find some time to work on it during the incoming days.

fendor:
I am not working on it right now.

fendor:
The first part of this proposal has been implemented, closing it in favour of #401

PR #271, opened 2020-12-16
Faster Bios protocol

pepeiborra:
This implements the proposal in #270 to obtain the OUTPUTS and DEPS in a single run.

The implementation is careful to preserve backwards compatibility. If both the new and old mechanisms are used, the new mechanism takes precedence.

### DIFF
--- a/README.md
+++ b/README.md
@@ -560,29 +560,23 @@ dependencies:
   - default.nix
 ```
 
-For the `Bios` cradle type, there is an optional field for specifying a program
-or shell command to obtain cradle dependencies from:
+For the `Bios` cradle type, the newline-separated cradle dependencies must be written out
+to the file specified by the `HIE_BIOS_DEPS` environment variable.
+
+Previous versions implemented a different mechanism for collecting cradle dependencies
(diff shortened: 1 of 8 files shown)


pepeiborra:
I would appreciate a fast turnaround to get this into Hackage as soon as possible.

fendor:
This does not require a major version bump, right?

We recently merged #264 but didn't release it yet. Would you be fine to be released together with #264 or should we make a release containing for this change? 

Will review the PR a couple of hours and I can do the release later today,

pepeiborra:
This on its own does not constitute a major version bump, since the changes are backwards compatible and hardly any impact for most users. E.g. I don't even need to make a ghcide release to use these changes, simply a Cabal/stack constraint.

Very happy to have it released today together with #264. Many thanks!!

jneira:
@fendor not sure if @wz1000 mentioned in irc there would be an issue with #264

jneira:
@fendor i am afraid i've not tested it, if it works for you it will be fine

fendor (approved):

PR #272, opened 2020-12-16
Prepare release 0.7.2

fendor:
cc @pepeiborra 

This MR is just to check CI is green after dropping #264


fendor:
Published: https://hackage.haskell.org/package/hie-bios-0.7.2

PR #274, opened 2020-12-26
Implement HIE_BIOS_GHC

### LINKED ISSUE #270: Proposal: faster Bios protocol

pepeiborra:
Closes #270

### DIFF
--- a/README.md
+++ b/README.md
@@ -369,9 +369,16 @@ There are two important environment variables:
 * `HIE_BIOS_ARG`: the source file that we want to load. Options returned by the `program` should be able to compile the given source file.
   * This environment variable is *only* available if a shell program is given.
 
+There are two additional environment variables:
+
+* `HIE_BIOS_DEPS`: the filepath to write the cradle dependencies
+* `HIE_BIOS_GHC`: the filepath to write the ghc path
+
 The program flow is roughly as follows:
 The process must consult the `HIE_BIOS_OUTPUT` environment variable and write a
-list of options to this file, separated by newlines. Once the process finishes
+list of options to this file, separated by newlines. Similarly, write a dependency
+per line out to the file specified by `HIE_BIOS_DEPS`, and the ghc path to the file
+specified by `HIE_BIOS_GHC`. Once the process finishes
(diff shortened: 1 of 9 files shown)


pepeiborra:
This is ready for review

adamse:
Looks like this also makes a nice solution for https://github.com/mpickering/hie-bios/issues/253

fendor:
Ill see that I review it tomorrow. 

So far the main changes are:

* remove `runGhcCmd` from `Cradle a`
* add `componentRunGhcCmd` for `ComponentOptions`
* add `HIE_BIOS_GHC` output variable for cradle actions.
* Adds some instances for `CradleLoadResult`

Did I catch most changes?

fendor:
I am worrying this breaks the HLS  wrapper.

Afaik, the flow is as follow:

* Find `hie.yaml`
* Use `runGhcCmd` to find the ghc version
* Launch the correct HLS version

How would that work with `componentRunGhcCmd`?

* Find `hie.yaml`
* Find some component?
* Use `componentRunGhcCmd` to find the ghc version
* Launch the correct HLS version

pepeiborra:
> How would that work with `componentRunGhcCmd`?

Presumably the wrapper is always invoked with an input file, right? So the flow would be

* Find `hie.yaml`
* Find the local cradle
* Load the file with the cradle, resulting in some `ComponentOptions`
* Use `componentRunGhcCmd` to find the ghc version
* Launch the correct HLS version

pepeiborra:
I assumed wrong - the wrapper is not invoked with an input file. I don't know how to solve this, it seems that tying the ghc libdir to a specific component is not a good idea after all

PR #276, opened 2021-01-08
Optparse-applicative CLI

teto:
Rather than adhoc solution. The help was listing unsupported commands (expand), this shouldn't happen anymore.


fendor (approved):
Awesome, thank you! LGTM!

teto:
glad I intend to senf follow up PRs so that users can increase verbositywithout recompiling ( as I did while debugging my issue.)

PR #277, opened 2021-01-11
configurable verbosity

teto:
some of the changes are just recommandations made by haskell-language-server, I saw it fit to make hie-bios follows the recommandation from the software it exists for.

I would like to make this change configurable on the CLI rather than hardcode it
https://github.com/mpickering/hie-bios/compare/master...teto:nix?expand=1#diff-268e4205229d03a6419924ecdf57a963794db10db21703878f4d22424f2c8b58R47  
Do you think its a good idea ? or maybe we could just remove the
`setVerbosity 0                       -- Set verbosity to zero just in case the user specified `-vx` in the options.` ?
The user can set -v0 himself ? and is there any reason not to show a detailed log in the first place ?

### DIFF
--- a/hie-bios.cabal
+++ b/hie-bios.cabal
@@ -1,6 +1,6 @@
 Cabal-Version:          2.2
 Name:                   hie-bios
-Version:                0.7.1
+Version:                0.7.2
 Author:                 Matthew Pickering <matthewtpickering@gmail.com>
 Maintainer:             Matthew Pickering <matthewtpickering@gmail.com>
 License:                BSD-3-Clause
--- a/exe/Main.hs
+++ b/exe/Main.hs
@@ -21,6 +21,11 @@ import Paths_hie_bios
 progVersion :: String
 progVersion = "hie-bios version " ++ showVersion version ++ " compiled by GHC " ++ cProjectVersion ++ "\n"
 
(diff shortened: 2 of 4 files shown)


fendor:
Please avoid unnecessary changes unless there is a real benefit. 
We do not follow hlint guidelines, so although this project is a foundation for hls, there is no reason to follow rules of some plugin, just because they are the default. (E.g., hie-bios existed when hls did not even have a hlint plugin)

hie-bios.cabal
```diff
@@ -1,6 +1,6 @@
 Cabal-Version:          2.2
 Name:                   hie-bios
-Version:                0.7.1
+Version:                0.7.2
```
fendor, line 3: Why this change?

teto: it's an error, gonna drop

teto:
sure for the hlint changes, gonna drop. Now I wonder about how to enable a more verbose hie-bios: should I:
1/  just disable the automatic `verbisotiy0` to let a cradle's `-v2` succeed 
2/  just let the user pass some arbitrary args that override a default `-v0`
3/ add an arugment `hie-bios --verbosity=1`

I assume we want the feature (I found it very helpful).

PR #278, opened 2021-01-11
Add a default .hlint.yaml file, disabling all hints currently triggered

ndmitchell:
Otherwise people might mistakenly try and send patches fixing them, e.g. #277. Also ensures they aren't shown in the IDE.

FWIW, the one that I would totally apply is:

```
src\HIE\Bios\Cradle.hs:637:26-53: Suggestion: Use catMaybes
Found:
  [comp | Just comp <- [mc]]
Perhaps:
  Data.Maybe.catMaybes [mc]
```

That seems much cleaner. Everything else has both pros and cons.

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -1,0 +1,80 @@
+# HLint configuration file
+# https://github.com/ndmitchell/hlint
+##########################
+
+# This file contains a template configuration file, which is typically
+# placed as .hlint.yaml in the root of your project
+
+
+# Warnings currently triggered by your code
+- ignore: {name: "Use newtype instead of data"}
+- ignore: {name: "Use camelCase"}
+- ignore: {name: "Redundant $"}
+- ignore: {name: "Redundant bracket"}
+- ignore: {name: "Functor law"}
+- ignore: {name: "Use catMaybes"}
+- ignore: {name: "Use <$>"}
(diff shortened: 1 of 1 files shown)


.hlint.yaml
```diff
@@ -0,0 +1,80 @@
...
+- ignore: {name: "Use camelCase"}
+- ignore: {name: "Redundant $"}
+- ignore: {name: "Redundant bracket"}
+- ignore: {name: "Functor law"}
+- ignore: {name: "Use catMaybes"}
+- ignore: {name: "Use <$>"}
+- ignore: {name: "Eta reduce"}
+- ignore: {name: "Use const"}
+- ignore: {name: "Move brackets to avoid $"}
+- ignore: {name: "Parse error: possibly incorrect indentation or mismatched brackets"}
```
fendor, line 20: What is that? Seems like a weird error to ignore 0.o

ndmitchell: I changed it to just ignore "Parse error", which is what I intended all along. The reason is tests\BiosTests.hs:389:1. If you evaluate that file without the right CPP set then you end up with a parse error. It might be reasonable to have an `#else` that gives you `nightly` or something, or to tell HLint to set some combination of CPP flags to make it take one variant of that branch, but as it stands, its a parse error.

PR #281, opened 2021-01-29
Add job to compile the windows wrapper in CI

fendor:
Since travis kind of stopped working, we compile the wrappers to find warnings and have a higher confidence that it is working


jneira (approved):
lgtm, thanks for take care of windows 😄

fendor:
We have to wait for GitHub actions, though, which are, for some reason, not triggered, right now

Issue #194, opened 2020-06-10
cabal wrapper uses always global ghc in PATH 


jneira:
* detected when testing ghcide in itself (see https://github.com/digital-asset/ghcide/issues/610)
* although you set the field `with-compiler: ghc-8.6.5` in the `cabal.project` hie-bios was using my default ghc in path (ghc-8.8.3)
```
PS D:\dev\ws\haskell\ghcide> ghc --version
The Glorious Glasgow Haskell Compilation System, version 8.8.3
PS D:\dev\ws\haskell\ghcide> ghcide --version
ghcide version: 0.2.0 (GHC: 8.6.5) (PATH: D:\bin\ghcide.exe)
PS D:\dev\ws\haskell\ghcide> cat .\cabal.project
packages: .

with-compiler: ghc-8.6.5
PS D:\dev\ws\haskell\ghcide> cabal clean
PS D:\dev\ws\haskell\ghcide> ghcide
ghcide version: 0.2.0 (GHC: 8.6.5) (PATH: D:\bin\ghcide.exe)
Ghcide setup tester in D:\dev\ws\haskell\ghcide.
Report bugs at https://github.com/digital-asset/ghcide/issues

Step 1/4: Finding files to test in D:\dev\ws\haskell\ghcide
Found 63 files

Step 2/4: Looking for hie.yaml files that control setup
Found 3 cradles

Step 3/4: Initializing the IDE

Step 4/4: Type checking the files
[INFO] Consulting the cradle for "D:\\dev\\ws\\haskell\\ghcide\\src\\Development\\IDE\\GHC\\WithDynFlags.hs"
> Resolving dependencies...
> Build profile: -w ghc-8.8.3 -O1
> In order, the following will be built (use -v for more details):
>  - ghcide-0.2.0 (lib) (first run)
> Configuring library for ghcide-0.2.0..
> Preprocessing library for ghcide-0.2.0..
    D:\dev\ws\haskell\ghcide\src\Development\IDE\GHC\WithDynFlags.hs
Hidden:   no
Range:    1:0-100001:0
Source:   compiler
Severity: DsError
Message:
  <command line>: cannot satisfy -package-id aeson-1.5.1.0-1e42a0cdbd4b065e4d6f1c804d63f35706230918:
  aeson-1.5.1.0-1e42a0cdbd4b065e4d6f1c804d63f35706230918 is unusable due to missing dependencies:
[...]

PR #282, opened 2021-02-01
Use the proper GHC version given by cabal

Kleidukos:
WIP with @fendor to use the proper GHC version provided by Cabal instead of using blindingly the one in $PATH.

### DIFF
--- a/src/HIE/Bios/Cradle.hs
+++ b/src/HIE/Bios/Cradle.hs
@@ -132,14 +132,12 @@ addCradleDeps deps c =
     addActionDeps :: CradleAction a -> CradleAction a
     addActionDeps ca =
       ca { runCradle = \l fp ->
-            runCradle ca l fp
-              >>= \case
-                CradleSuccess (ComponentOptions os' dir ds) ->
-                  pure $ CradleSuccess (ComponentOptions os' dir (ds `union` deps))
-                CradleFail err ->
-                  pure $ CradleFail
-                    (err { cradleErrorDependencies = cradleErrorDependencies err `union` deps })
-                CradleNone -> pure CradleNone
+            (cradleLoadResult
+                CradleNone
+                (\err -> CradleFail (err { cradleErrorDependencies = cradleErrorDependencies err `union` deps }))
+                (\(ComponentOptions os' dir ds) -> CradleSuccess (ComponentOptions os' dir (ds `union` deps)))
+            )
+            <$> runCradle ca l fp
          }
 
 -- | Try to infer an appropriate implicit cradle type from stuff we can find in the enclosing directories:
@@ -431,16 +429,15 @@ cabalCradle wdir mc =
     { cradleRootDir    = wdir
     , cradleOptsProg   = CradleAction
         { actionName = Types.Cabal
-        , runCradle = cabalAction wdir mc
-        , runGhcCmd = \args -> do
-            buildDir <- cabalBuildDir wdir
+        , runCradle = \l f -> runCradleResultT . cabalAction wdir mc l $ f
+        , runGhcCmd = \args -> runCradleResultT $ do
+            buildDir <- liftIO $ cabalBuildDir wdir
             -- Workaround for a cabal-install bug on 3.0.0.0:
             -- ./dist-newstyle/tmp/environment.-24811: createDirectory: does not exist (No such file or directory)
-            createDirectoryIfMissing True (buildDir </> "tmp")
+            liftIO $ createDirectoryIfMissing True (buildDir </> "tmp")
             -- Need to pass -v0 otherwise we get "resolving dependencies..."
-            cabalProcLoadResult <- cabalProcess wdir "v2-exec" $ ["ghc", "-v0", "--"] ++ args
-            cabalProcLoadResult `bindIO` \cabalProc ->
-              readProcessWithCwd' cabalProc ""
+            cabalProc <- cabalProcess wdir "v2-exec" $ ["ghc", "-v0", "--"] ++ args
+            readProcessWithCwd' cabalProc ""
         }
     }
 
@@ -453,16 +450,15 @@ cabalCradle wdir mc =
 -- to the custom ghc wrapper via 'HIE_BIOS_GHC' environment variable which
 -- the custom ghc wrapper may use as a fallback if it can not respond to certain
 -- queries, such as ghc version or location of the libdir.
-cabalProcess :: FilePath -> String -> [String] -> IO (CradleLoadResult CreateProcess)
+cabalProcess :: FilePath -> String -> [String] -> CradleLoadResultT IO CreateProcess
 cabalProcess workDir command args = do
-  ghcDirLoadResult <- cabalGhcDirs workDir
-  ghcDirLoadResult `bindIO` \ghcDirs -> do
-    newEnvironment <- setupEnvironment ghcDirs
-    cabalProc <- setupCabalCommand ghcDirs
-    pure $ CradleSuccess (cabalProc
-        { env = Just newEnvironment
-        , cwd = Just workDir
-        })
+  ghcDirs <- cabalGhcDirs workDir
+  newEnvironment <- liftIO $ setupEnvironment ghcDirs
+  cabalProc <- liftIO $ setupCabalCommand ghcDirs
+  pure $ (cabalProc
+      { env = Just newEnvironment
+      , cwd = Just workDir
+      })
   where
     processEnvironment :: (FilePath, FilePath) -> [(String, String)]
     processEnvironment (ghcBin, libdir) =
@@ -648,42 +644,36 @@ cabalBuildDir workDir = do
 -- 'withGhcWrapperTool'.
 --
 -- If cabal can not figure it out, a 'CradleError' is returned.
-cabalGhcDirs :: FilePath -> IO (CradleLoadResult (FilePath, FilePath))
+cabalGhcDirs :: FilePath -> CradleLoadResultT IO (FilePath, FilePath)
 cabalGhcDirs workDir = do
-  libdirLoadResult <- readProcessWithCwd workDir  "cabal" ["exec", "-v0", "--", "ghc", "--print-libdir"] ""
-  libdirLoadResult `bindIO` \libdir -> do
-    exePathLoadResult <- readProcessWithCwd workDir  "cabal"
-        ["exec", "-v0", "--", "ghc", "-package-env=-", "-e", "do e <- System.Environment.getExecutablePath ; System.IO.putStr e"] ""
-    exePathLoadResult `bindIO` \exe -> pure $ CradleSuccess (trimEnd exe, trimEnd libdir)
+  libdir <- readProcessWithCwd_ workDir  "cabal" ["exec", "-v0", "--", "ghc", "--print-libdir"] ""
+  exe <- readProcessWithCwd_ workDir  "cabal"
+      ["exec", "-v0", "--", "ghc", "-package-env=-", "-e", "do e <- System.Environment.getExecutablePath ; System.IO.putStr e"] ""
+  pure (trimEnd exe, trimEnd libdir)
 
-cabalAction :: FilePath -> Maybe String -> LoggingFunction -> FilePath -> IO (CradleLoadResult ComponentOptions)
-cabalAction workDir mc l fp =
-  cabalProcess workDir "v2-repl" [fromMaybe (fixTargetPath fp) mc] >>= \case
-    CradleNone -> pure CradleNone
-    CradleFail err -> do
+cabalAction :: FilePath -> Maybe String -> LoggingFunction -> FilePath -> CradleLoadResultT IO ComponentOptions
+cabalAction workDir mc l fp = do
+  cabalProc <- (cabalProcess workDir "v2-repl" [fromMaybe (fixTargetPath fp) mc]) `modCradleError` \err -> do
+      deps <- cabalCradleDependencies workDir workDir
+      pure $ err { cradleErrorDependencies = cradleErrorDependencies err ++ deps }
+
+  (ex, output, stde, [(_, maybeArgs)]) <- liftIO $ readProcessWithOutputs [hie_bios_output] l workDir cabalProc
+
+  let args = fromMaybe [] maybeArgs
+  case processCabalWrapperArgs args of
+    Nothing -> do
       -- Provide some dependencies an IDE can look for to trigger a reload.
       -- Best effort. Assume the working directory is the
       -- root of the component, so we are right in trivial cases at least.
-      deps <- cabalCradleDependencies workDir workDir
-      pure $ CradleFail err { cradleErrorDependencies = cradleErrorDependencies err ++ deps }
-    CradleSuccess cabalProc -> do
-      (ex, output, stde, [(_, maybeArgs)]) <- readProcessWithOutputs [hie_bios_output] l workDir cabalProc
-
-      let args = fromMaybe [] maybeArgs
-      case processCabalWrapperArgs args of
-        Nothing -> do
-          -- Provide some dependencies an IDE can look for to trigger a reload.
-          -- Best effort. Assume the working directory is the
-          -- root of the component, so we are right in trivial cases at least.
-          deps <- cabalCradleDependencies workDir workDir
-          pure $ CradleFail (CradleError deps ex
-                    ["Failed to parse result of calling cabal"
-                    , unlines output
-                    , unlines stde
-                    , unlines $ args])
-        Just (componentDir, final_args) -> do
-          deps <- cabalCradleDependencies workDir componentDir
-          pure $ makeCradleResult (ex, stde, componentDir, final_args) deps
+      deps <- liftIO $ cabalCradleDependencies workDir workDir
+      throwCE (CradleError deps ex
+                ["Failed to parse result of calling cabal"
+                , unlines output
+                , unlines stde
+                , unlines $ args])
+    Just (componentDir, final_args) -> do
+      deps <- liftIO $ cabalCradleDependencies workDir componentDir
+      CradleLoadResultT $ pure $ makeCradleResult (ex, stde, componentDir, final_args) deps
   where
     -- Need to make relative on Windows, due to a Cabal bug with how it
     -- parses file targets with a C: drive in it
@@ -764,14 +754,13 @@ stackCradle wdir mc syaml =
     , cradleOptsProg   = CradleAction
         { actionName = Types.Stack
         , runCradle = stackAction wdir mc syaml
-        , runGhcCmd = \args -> do
+        , runGhcCmd = \args -> runCradleResultT $ do
             -- Setup stack silently, since stack might print stuff to stdout in some cases (e.g. on Win)
             -- Issue 242 from HLS: https://github.com/haskell/haskell-language-server/issues/242
-            stackSetup <- readProcessWithCwd wdir "stack" (stackYamlProcessArgs syaml <> ["setup", "--silent"]) ""
-            stackSetup `bindIO` \_ ->
-              readProcessWithCwd wdir "stack"
-                (stackYamlProcessArgs syaml <> ["exec", "ghc", "--"] <> args)
-                ""
+            _ <- readProcessWithCwd_ wdir "stack" (stackYamlProcessArgs syaml <> ["setup", "--silent"]) ""
+            readProcessWithCwd_ wdir "stack"
+              (stackYamlProcessArgs syaml <> ["exec", "ghc", "--"] <> args)
+              ""
         }
     }
 
@@ -1036,22 +1025,25 @@ runGhcCmdOnPath wdir args = readProcessWithCwd wdir "ghc" args ""
 -- | Wrapper around 'readCreateProcess' that sets the working directory and
 -- clears the environment, suitable for invoking cabal/stack and raw ghc commands.
 readProcessWithCwd :: FilePath -> FilePath -> [String] -> String -> IO (CradleLoadResult String)
-readProcessWithCwd dir cmd args stdin = do
-  cleanEnv <- getCleanEnvironment
+readProcessWithCwd dir cmd args stdin = runCradleResultT $ readProcessWithCwd_ dir cmd args stdin
+
+readProcessWithCwd_ :: FilePath -> FilePath -> [String] -> String -> CradleLoadResultT IO String
+readProcessWithCwd_ dir cmd args stdin = do
+  cleanEnv <- liftIO getCleanEnvironment
   let createProc = (proc cmd args) { cwd = Just dir, env = Just cleanEnv }
   readProcessWithCwd' createProc stdin
 
 -- | Wrapper around 'readCreateProcessWithExitCode', wrapping the result in
 -- a 'CradleLoadResult'. Provides better error messages than raw 'readCreateProcess'.
-readProcessWithCwd' :: CreateProcess -> String -> IO (CradleLoadResult String)
+readProcessWithCwd' :: CreateProcess -> String -> CradleLoadResultT IO String
 readProcessWithCwd' createdProcess stdin = do
-  mResult <- optional $ readCreateProcessWithExitCode createdProcess stdin
+  mResult <- liftIO $ optional $ readCreateProcessWithExitCode createdProcess stdin
   let cmdString = prettyCmdSpec $ cmdspec createdProcess
   case mResult of
-    Just (ExitSuccess, stdo, _) -> pure $ CradleSuccess stdo
-    Just (exitCode, stdo, stde) -> pure $ CradleFail $
+    Just (ExitSuccess, stdo, _) -> pure stdo
+    Just (exitCode, stdo, stde) -> throwCE $
       CradleError [] exitCode ["Error when calling " <> cmdString, stdo, stde]
-    Nothing -> pure $ CradleFail $
+    Nothing -> throwCE $
       CradleError [] ExitSuccess ["Couldn't execute " <> cmdString]
 
 -- | Prettify 'CmdSpec', so we can show the command to a user
--- a/src/HIE/Bios/Cradle.hs
+++ b/src/HIE/Bios/Cradle.hs
@@ -1,7 +1,8 @@
+{-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TupleSections #-}
-{-# LANGUAGE BangPatterns #-}
 module HIE.Bios.Cradle (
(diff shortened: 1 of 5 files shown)


fendor:
closes #194

src/HIE/Bios/Cradle.hs
```diff
@@ -493,37 +493,55 @@ withCabalWrapperTool (mbGhc, ghcArgs) wdir = do
...
 
+cabalGHCBin :: IO FilePath
+cabalGHCBin = do
+  result <- readProcessWithCwd "." "cabal" ["exec", "-v0", "--", "ghc", "--print-libdir"] ""
+  case result of
+    CradleFail err -> error (show err)
+    CradleNone     -> undefined
+    CradleSuccess r -> do
+      let strippedPath = T.unpack $ T.stripEnd $ T.pack r
+      let cabalGHC = strippedPath </> "../../bin/ghc"
```
jneira, line 509: It seems the path in windows is different (only one `..`)?

```
PS D:\hls> cabal exec ghc -- --print-libdir
D:\bin\stack\x86_64-windows\ghc-8.8.4\lib
PS D:\hls> D:\bin\stack\x86_64-windows\ghc-8.8.4\lib\..\bin\ghc --version
The Glorious Glasgow Haskell Compilation System, version 8.8.4
```

fendor: well that is certainly bad and answers whether the libdir is always relative to the ghc binary location

src/HIE/Bios/Cradle.hs
```diff
@@ -413,12 +413,38 @@ cabalCradle wdir mc =
...
+-- | Execute a cabal process in our custom cache-build directory configured
+-- with the custom ghc executable.
+-- The created process has its working directory set to the given working directory.
+--
+-- Invokes the cabal process in the given directory.
+-- Finds the appropriate @ghc@ version as a fallback and provides the path
+-- to the custom ghc wrapper via 'HIE_BIOS_GHC' environment variable which
+-- the custom ghc wrapper may use as a fallback if it can not respond to certain
+-- queries, such as ghc version or location of the libdir.
+cabalProcess :: FilePath -> String -> [String] -> IO (CradleLoadResult CreateProcess)
```
fendor, line 431: Not sure if this is a good API, but since it is not exposed, we can change it any time

src/HIE/Bios/Cradle.hs
```diff
@@ -493,37 +519,57 @@ withCabalWrapperTool (mbGhc, ghcArgs) wdir = do
...
+cabalGHCDir :: FilePath -> IO (CradleLoadResult FilePath)
+cabalGHCDir workDir = do
+  readProcessWithCwd workDir  "cabal" ["exec", "-v0", "--", "ghc", "--print-libdir"] ""
+    >>= \case
+      CradleSuccess r -> do
+        let strippedPath = T.unpack $ T.stripEnd $ T.pack r
+        pure $ CradleSuccess $
+          if isWindows
+            then strippedPath </>"../bin"
+            else strippedPath </> "../../bin"
```
fendor, line 536: I suggested that this should be a recursive search function, e.g. look in every parent directory whether there is a `bin` and see if it contains what we need. 
Your opinion @jneira, or whoever read this?

src/HIE/Bios/Cradle.hs
```diff
@@ -493,37 +519,57 @@ withCabalWrapperTool (mbGhc, ghcArgs) wdir = do
...
+cabalAction workDir mc l fp = do
+    cabalProcess workDir "v2-repl" [fromMaybe (fixTargetPath fp) mc]
+     >>= \case
+      CradleNone -> pure CradleNone
+      CradleFail err -> do
+        -- Provide some dependencies an IDE can look for to trigger a reload.
+        -- Best effort. Assume the working directory is the
+        -- root of the component, so we are right in trivial cases at least.
+        deps <- cabalCradleDependencies workDir workDir
+        pure $ CradleFail err { cradleErrorDependencies = cradleErrorDependencies err ++ deps }
```
fendor, line 549: Somewhat code duplication, but hard to avoid and local

src/HIE/Bios/Types.hs
```diff
@@ -74,8 +76,25 @@ data CradleLoadResult r
...
+instance Monad CradleLoadResult where
+  return = CradleSuccess
+  CradleSuccess r >>= k = k r
+  CradleFail err >>= _ = CradleFail err
+  CradleNone >>= _ = CradleNone
+
+bindIO :: CradleLoadResult a -> (a -> IO (CradleLoadResult b)) -> IO (CradleLoadResult b)
+bindIO  (CradleSuccess r) k = k r
+bindIO (CradleFail err) _ = return $ CradleFail err
+bindIO CradleNone _ = return CradleNone
```
fendor, line 97: Shamelessly stolen from https://github.com/mpickering/hie-bios/pull/274/files#diff-f173b250a7067c10436359a14249f8e3386648a9d009c44272ffc0c7d758c03cR75 since I really liked that. @pepeiborra Should we attribute it to you somehow?

pepeiborra: No need

src/HIE/Bios/Cradle.hs
jneira: ghc itself [does a recursive function with a max depth of three](https://gitlab.haskell.org/ghc/ghc/-/blob/master/compiler/GHC/SysTools/BaseDir.hs#L184-203) to discover its mingw internal 😄 

```haskell
findToolDir
  :: FilePath -- ^ topdir
  -> IO (Maybe FilePath)
#if defined(mingw32_HOST_OS) && !defined(USE_INPLACE_MINGW_TOOLCHAIN)
findToolDir top_dir = go 0 (top_dir </> "..") []
  where maxDepth = 3
        go :: Int -> FilePath -> [FilePath] -> IO (Maybe FilePath)
        go k path tried
          | k == maxDepth = throwGhcExceptionIO $
              InstallationError $ "could not detect mingw toolchain in the following paths: " ++ show tried
          | otherwise = do
              let try = path </> "mingw"
              let tried = tried ++ [try]
              oneLevel <- doesDirectoryExist try
              if oneLevel
                then return (Just path)
                else go (k+1) (path </> "..") tried
#else
findToolDir _ = return Nothing
#endif
```
so this version is fine as is, i guess

jneira (approved):
many thanks @fendor and @Kleidukos !

jneira:
This can be merged, no? It will be included in the next release though

fendor:
I am having doubts about the implementation. In particular, I am worried we might break everything irreparable for expert users in non-standard systems.

jneira:
> 
> 
> I am having doubts about the implementation. In particular, I am worried we might break everything irreparable for expert users in non-standard systems.

Are you thinking in nix, f.e.? what kind of problems could this cause?

fendor:
No, I am just thinking of people who supply their libdir path to ghc directly and use a non-expected relative location. We won't find the ghc executable in that case, hie-bios will not find anything, thus, hls will stop working for those people, although the correct `ghc` might be on the path.

So, as a work-around, if we don't find the ghc location, we default to the previous behaviour? Then we only improve the situation for some without destroying it for others?

jneira:
> as a work-around, if we don't find the ghc location, we default to the previous behaviour?

well, looking for ghc in path, or even better, ghc-${version} if we can extract the version from `cabal exec ghc -- --version` could be the last resort for all ghc lookups (cabal, stack, etc, etc)

jneira:
I am thinking in a ugly hack:

```
D:\ws\haskell\cabal-test>cabal exec ghc -- -v0 -e "System.Environment.getExecutablePath"
"D:\\bin\\Programs\\stack\\x86_64-windows\\ghc-8.8.4\\bin\\ghc.exe"
```

:laughing: :rofl: :relaxed:

jneira:
@fendor could we try to revisit this pr and merge it?

I think we have to take a decision about the fallback

> So, as a work-around, if we don't find the ghc location, we default to the previous behaviour? Then we only improve the situation for some without destroying it for others?

I think we could start with this and go ahead

fendor:
@jneira Thank you for reminding us of these issue, will spend some brain cells on it!

fendor:
As with everything in cabal, doesn't work on first try. `cabal exec -- ghc -e "putStrLn =<< System.Environment.getExecutablePath"` fails for me with 
```
ghc: can't find a package database at /home/hugin/Documents/haskell/hie-bios/dist-newstyle/packagedb/ghc-8.10.4
```
if we haven't invoked `cabal build` before.

Maybe it would make sense to call `cabal build --only-configure` or something like that before?

jneira:
Friendly ping, an user reported this in irc

jneira:
ha! it seems `cabal exec ghc -- -package-env=- -e 'System.Environment.getExecutablePath'` 

```
PS D:\dev\ws\haskell\hie-bios> rmdir .\dist-newstyle\ -rec -for
PS D:\dev\ws\haskell\hie-bios> cabal exec ghc -- -package-env=- -e 'System.Environment.getExecutablePath'
Cloning into 'D:\dev\ws\haskell\hie-bios\dist-newstyle\src\ghc-api-c_-c0de97d0472b239c'...
remote: Enumerating objects: 108, done.
remote: Counting objects: 100% (70/70), done.
remote: Compressing objects: 100% (47/47), done.
remote: Total 108 (delta 41), reused 42 (delta 20), pack-reused 38
Receiving objects: 100% (108/108), 40.23 KiB | 633.00 KiB/s, done.
Resolving deltas: 100% (59/59), done.
HEAD is now at 6178d75 Add base upper bound
Resolving dependencies...
"D:\\ghcup\\ghc\\9.0.1\\bin\\ghc.exe"
```

hasufell:
It appears this should work fine:

1. get libdir via `cabal exec ghc -- --print-libdir` 
2. get binary via `cabal exec ghc -- -package-env=- -e 'putStrLn =<< System.Environment.getExecutablePath'`
3. create a shim (or adjust the one that already exists): `/full/path/to/ghc -B$libdir "$@"`
4. profit

src/HIE/Bios/Cradle.hs
```diff
@@ -493,37 +527,52 @@ withCabalWrapperTool (mbGhc, ghcArgs) wdir = do
...
+  abs_work_dir <- makeAbsolute workDir
   let dirHash = show (fingerprintString abs_work_dir)
-  getCacheDir ("dist-"<>filter (not . isSpace) (takeBaseName abs_work_dir)<>"-"<>dirHash)
+  getCacheDir ("dist-" <> filter (not . isSpace) (takeBaseName abs_work_dir)<>"-"<>dirHash)
+
+cabalGHCDir :: FilePath -> IO (CradleLoadResult (FilePath, FilePath))
+cabalGHCDir workDir = do
+  libdirLoadResult <- readProcessWithCwd workDir  "cabal" ["exec", "-v0", "--", "ghc", "--print-libdir"] ""
+  libdirLoadResult `bindIO` \libdir -> do
+    exePathLoadResult <- readProcessWithCwd workDir  "cabal" ["exec", "-v0", "--", "ghc", "-e", "putStrLn =<< System.Environment.getExecutablePath"] ""
```
hasufell, line 539: This actually needs to be:

```
do e <- System.Environment.getExecutablePath ; System.IO.putStr e
```

in case someone has ImplicitPrelude disables globally/locally.

See https://github.com/haskell/haskell-language-server/pull/2711

src/HIE/Bios/Cradle.hs
```diff
@@ -519,27 +563,30 @@ type GhcProc = (FilePath, [String])
...
-    let wrapper_name = "wrapper-" ++ srcHash
-    let wrapper_fp = suffix $ cacheDir </> wrapper_name
-    exists <- doesFileExist wrapper_fp
-    unless exists $
+    cacheFile (withExtension "wrapper") srcHash $ \wrapper_fp ->
       if isWindows
       then
         withSystemTempDirectory "hie-bios" $ \ tmpDir -> do
-        let wrapper_hs = cacheDir </> wrapper_name <.> "hs"
+        let wrapper_hs = (takeDirectory wrapper_fp) </> (takeFileName wrapper_fp) -<.> "hs"
```
fendor, line 573:
```suggestion
        let wrapper_hs = wrapper_fp -<.> "hs"
```

src/HIE/Bios/Cradle.hs
```diff
@@ -466,13 +470,53 @@ cabalProcess workDir command args = do
...
+      case parseCompilerId (T.pack ghcName) of
+        Nothing -> ghcDir </> "ghc-pkg"
+        Just version -> ghcDir </> "ghc-pkg-" <> showVersion version
+
+    withWrapperTool ghcPkg = do
+      let globalPackageDb = libdir </> "package.conf.d"
+      let contents = unlines
+            [ "#!/bin/sh"
+            , unwords ["command", ghcPkg, "--global-package-db", globalPackageDb, "${1+\"$@\"}"]
+            ]
```
fendor, line 506: Should this be documented, or extraced just like the cabal wrapper?

src/HIE/Bios/Cradle.hs
```diff
@@ -21,18 +21,21 @@ module HIE.Bios.Cradle (
     , readProcessWithOutputs
     , readProcessWithCwd
     , makeCradleResult
+    , withGhcPkgTool
```
fendor, line 24: Don't export this function

src/HIE/Bios/Cradle.hs
```diff
@@ -466,13 +470,53 @@ cabalProcess workDir command args = do
...
+  let ghcPkgPath = guessGhcPkgFromGhc ghcName
+  if isWindows
+    then pure ghcPkgPath
+    else withWrapperTool ghcPkgPath
+  where
+    ghcDir = takeDirectory ghcPathAbs
+
+    guessGhcPkgFromGhc ghcName =
+      case parseCompilerId (T.pack ghcName) of
+        Nothing -> ghcDir </> "ghc-pkg"
```
hasufell, line 498: Does this work on windows without exe extension?

src/HIE/Bios/Cradle.hs
hasufell: Def. docs

src/HIE/Bios/Cradle.hs
```diff
@@ -466,13 +470,53 @@ cabalProcess workDir command args = do
...
+        Just version -> ghcDir </> "ghc-pkg-" <> showVersion version
+
+    withWrapperTool ghcPkg = do
+      let globalPackageDb = libdir </> "package.conf.d"
+      let contents = unlines
+            [ "#!/bin/sh"
+            , unwords ["command", ghcPkg, "--global-package-db", globalPackageDb, "${1+\"$@\"}"]
```
hasufell, line 505: command? Shouldn't this be exec?

src/HIE/Bios/Cradle.hs
```diff
@@ -466,13 +470,53 @@ cabalProcess workDir command args = do
...
+            ]
+      let srcHash = show (fingerprintString contents)
+      cacheFile "ghc-pkg" srcHash $ \wrapperFp -> writeFile wrapperFp contents
+
+    parseInnerVersion :: T.Text -> Maybe Version
```
hasufell, line 510: I think parse will fail on windows

src/HIE/Bios/Cradle.hs
```diff
@@ -466,13 +470,53 @@ cabalProcess workDir command args = do
...
+          pure ["--with-hc-pkg", ghcPkgPath]
+        _ -> pure []
+  let cabalProc = proc "cabal" (baseCabalArgs ++ extraCabalArgs ++ args)
   pure $ CradleSuccess (cabalProc
       { env = Just newEnvironment
       , cwd = Just workDir
       })
 
+withGhcPkgTool :: FilePath -> FilePath -> IO FilePath
+withGhcPkgTool ghcPathAbs libdir = do
```
hasufell, line 486: Document our assumptions here and why it's different on windows, maybe

src/HIE/Bios/Cradle.hs
```diff
@@ -466,13 +470,53 @@ cabalProcess workDir command args = do
...
+  -- TODO: check for existence
+  let ghcPkgPath = guessGhcPkgFromGhc ghcName
+  if isWindows
+    then pure ghcPkgPath
+    else withWrapperTool ghcPkgPath
+  where
+    ghcDir = takeDirectory ghcPathAbs
+
+    guessGhcPkgFromGhc ghcName =
+      case parseCompilerId (T.pack ghcName) of
```
hasufell, line 497: Maybe just strip the `ghc` prefix and prepend `ghc-pkg`? Then we can omit the parsing I guess.

src/HIE/Bios/Cradle.hs
fendor: This wrapper is only executed on the non-windows code path, so, hopefully

fendor: Should be exec (no idea why I wrote command)

src/HIE/Bios/Cradle.hs
fendor: What should happen in the 'Nothing' case?

src/HIE/Bios/Cradle.hs
```diff
@@ -435,12 +437,108 @@ cabalCradle wdir mc =
...
+  extraCabalArgs <- case ghcDirLoadResult of
+        CradleSuccess (ghcBin, libdir) -> do
+          ghcPkgPath <- withGhcPkgTool ghcBin libdir
+          pure ["--with-hc-pkg", ghcPkgPath]
+        _ -> pure []
+  let cabalProc = proc "cabal" (baseCabalArgs ++ extraCabalArgs ++ args)
+  pure $ CradleSuccess (cabalProc
+      { env = Just newEnvironment
+      , cwd = Just workDir
+      })
```
fendor, line 480: TODO: refactor this messy function

src/HIE/Bios/Cradle.hs
hasufell: `fail` hard

hasufell: or wait... it's possible that distros use arch triple prefix, e.g. `x86_64-redhat-linux-ghc`... even ghcup supports prefixed cross GHCs, so the most robust way would be to just `s/ghc/ghc-pkg/`... not assuming prefix.

hasufell: On the other hand:

```
echo 'exeprog="ghc-stage1"'                          >> "/builds/haskell/ghcup-hs/.ghcup/ghc/arm-linux-gnueabihf-8.10.7/bin/arm-linux-gnueabihf-ghc-8.10.7"
echo 'executablename="$exedir/$exeprog"'           >> "/builds/haskell/ghcup-hs/.ghcup/ghc/arm-linux-gnueabihf-8.10.7/bin/arm-linux-gnueabihf-ghc-8.10.7"
echo 'datadir="/builds/haskell/ghcup-hs/.ghcup/ghc/arm-linux-gnueabihf-8.10.7/share"'                             >> "/builds/haskell/ghcup-hs/.ghcup/ghc/arm-linux-gnueabihf-8.10.7/bin/arm-linux-gnueabihf-ghc-8.10.7"
echo 'bindir="/builds/haskell/ghcup-hs/.ghcup/ghc/arm-linux-gnueabihf-8.10.7/bin"'                               >> "/builds/haskell/ghcup-hs/.ghcup/ghc/arm-linux-gnueabihf-8.10.7/bin/arm-linux-gnueabihf-ghc-8.10.7"
echo 'topdir="/builds/haskell/ghcup-hs/.ghcup/ghc/arm-linux-gnueabihf-8.10.7/lib/arm-linux-gnueabihf-ghc-8.10.7"'                               >> "/builds/haskell/ghcup-hs/.ghcup/ghc/arm-linux-gnueabihf-8.10.7/bin/arm-linux-gnueabihf-ghc-8.10.7"
echo 'executablename="$exedir/ghc"' >> "/builds/haskell/ghcup-hs/.ghcup/ghc/arm-linux-gnueabihf-8.10.7/bin/arm-linux-gnueabihf-ghc-8.10.7"
cat ghc/ghc.wrapper                         >> "/builds/haskell/ghcup-hs/.ghcup/ghc/arm-linux-gnueabihf-8.10.7/bin/arm-linux-gnueabihf-ghc-8.10.7"
```

I don't have the actual tarball at hand, but this indicates that for a cross ghc, the actual binary name is `ghc-stage1`. This supports my theory that we shouldn't parse the version. Not assuming a prefix might make it additionally more robust.

fendor: Replace sounds like a good idea.

fendor: `replace` sounds fine. Changed it, although what happens if "ghc" doesn't occur in the name? Should we check for that case as well?

src/HIE/Bios/Cradle.hs
fendor: Docs are improved now.

wrappers/cabal
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -o xtrace
+
```
fendor, line 4: Is it safe to keep this in? Or will we add more undesired logs to HLS?

fendor (changes requested):

fendor (approved):
LGTM, let's push this forward (lol, of course my approval doesn't matter in this case)

wrappers/cabal
hasufell: it's safe in the sense that it's specified by posix: https://pubs.opengroup.org/onlinepubs/007904875/utilities/set.html

Whether it's desired, I don't know.

hasufell: I just checked... it spams the output really hard: https://paste.tomsmeding.com/WllOnqGK

fendor: Ah thank you! Going to remove it then.

src/HIE/Bios/Cradle.hs
```diff
@@ -435,12 +437,107 @@ cabalCradle wdir mc =
...
+    -- a look at @cat $(which ghc-pkg)@, assuming 'ghc-pkg' is on your $PATH.
+    --
+    -- If we used the raw executable, i.e. not wrapped in a shim, then 'cabal'
+    -- can not use the given 'ghc-pkg'.
+    withWrapperTool ghcPkg = do
+      let globalPackageDb = libdir </> "package.conf.d"
+          -- This is the same as the wrapper-shims ghc-pkg usually comes with.
+          contents = unlines
+            [ "#!/bin/sh"
+            , unwords ["exec", ghcPkg, "--global-package-db", globalPackageDb, "${1+\"$@\"}"]
```
hasufell, line 536:
```suggestion
            , unwords ["exec", "\"" <> ghcPkg <> "\"", "--global-package-db", "\"" <> globalPackageDb <> "\"", "${1+\"$@\"}"]
```

Here we are missing quotes for both `ghcPkg` and `globalPackageDb`

The resulting wrapper script looks like this:

```sh
#!/bin/sh
exec /home/hasufell/.ghcup/ghc/8.8.4/lib/ghc-8.8.4/bin/ghc-pkg --global-package-db /home/hasufell/.ghcup/ghc/8.8.4/lib/ghc-8.8.4/package.conf.d ${1+"$@"}
```

src/HIE/Bios/Cradle.hs
```diff
@@ -483,69 +579,109 @@ processCabalWrapperArgs args =
...
-      then do
+      then
         withSystemTempDirectory "hie-bios" $ \ tmpDir -> do
-          let wrapper_hs = cacheDir </> wrapper_name <.> "hs"
+          let wrapper_hs = wrapper_fp -<.> "hs"
           writeFile wrapper_hs wrapperContents
           let ghc = (proc mbGhc $
                       ghcArgs ++ ["-rtsopts=ignore", "-outputdir", tmpDir, "-o", wrapper_fp, wrapper_hs])
                       { cwd = Just wdir }
           readCreateProcess ghc "" >>= putStr
```
hasufell, line 599: What is this for, btw?

fendor: We basically need a shim for ghc on windows to intercept the ghc arguments cabal would pass to ghc. However, there is no simple shim available, thus we compile a hs file that is the shim.

src/HIE/Bios/Cradle.hs
fendor: since unix filepaths can contain `"`, we should call `show` on our strings, right?

wz1000: Conisder [`GHC.ResponseFile.escapeArgs`](https://hackage.haskell.org/package/base-4.14.1.0/docs/GHC-ResponseFile.html#v:escapeArgs)

hasufell:
I find the `bindIO` rather annoying, check out https://github.com/Kleidukos/hie-bios/pull/1 introducing a `CradleLoadResultT`.

src/HIE/Bios/Cradle.hs
hasufell: > `replace` sounds fine. Changed it, although what happens if "ghc" doesn't occur in the name? Should we check for that case as well?

we could just let the world descend into chaos

fendor:
@jneira would you mind blow testing whether this works as intended on windows? I tried to come up with some tests, but it complicates CI quite a bit as we have to install two ghc versions, afaict

src/HIE/Bios/Cradle.hs
fendor: escapeArgs adds newlines, which I have trimmed away. I think this also might trim away other potentially correct paths from a filepath, but I think it suffices to patch it *if* it breaks on someones system

fendor:
oh well, I believe we can drop 8.4.4 support, since we have, afaik, no other consumer than HLS which supports only 8.8.4. So we can even drop support for 8.6.5 as well.

jneira:
> @jneira would you mind blow testing whether this works as intended on windows? 

I am giving a try and i am getting hangs, i am still investigating what could be the cause. Not sure if it would be a local issue.
Watching processes i check how hie-bios calls cabal, which in turn call ghc, cc, etc. At some point hie-bios and cabal hangs.

jneira:
> So we can even drop support for 8.6.5 as well.

hmm hls still supports 8.6.5, no?

fendor:
> hmm hls still supports 8.6.5, no?

My mistake, don't know why I thought we dropped support

jneira:
> it complicates CI quite a bit as we have to install two ghc versions, afaict

I think the machinery to accomplish the feature has become complex enough to setup such test, it will be good for all os's.
In theory you have to call haskell-setup two times. But i think gihub vms have several ghcs preinstalled (three last major versions?, so chosen the right ones will not take so much time.

fendor:
Ok, I wrote a test, windows looks good, only thing missing is dropping 8.4 support.

Any objections or other todos? @jneira, @Kleidukos, @hasufell

hasufell:
Let's go

jneira (approved):
Sure, great work, congrats all

PR #285, opened 2021-02-22
ghc-9.0.1 support

jneira:
* Using ghc-api-compat head
* The hard work was done by @fendor
* Only 2 libs with allow-newer!


jneira:
ghc-api-compat does not suport ghc-8.4, time to drop support for it?

fendor:
I would rather try to attempt adding ghc-8.4 support to ghc-api-compat. I mean, how hard can that be? (famous last words)

anka-213:
Does any downstream consumers of this library support ghc-8.4? 
Haskell Language Server/ghcide only supports ghc-8.6 and up, and judging from this list, there are not that many other users: https://packdeps.haskellers.com/reverse/hie-bios

fendor:
I am honestly not aware of any other consumers of this library

anka-213:
Should we just drop ghc-8.4 support then?

fendor:
If no one objects, sure

jneira:
Superseded by #300

PR #287, opened 2021-02-24
Pass dynamic flag for dynamic GHC

berberman:
We need reflect the dynamic flag to default cradle and direct cradle, otherwise users are required to write an explicit hie.yaml to pass the dynamic flag. We don't  want to pass the flag to stack cradle or cabal cradle, since I think we can assume that users have already configured the dynamic flag manually in their build tools.

By adding this, most of ghcide's tests can pass on Arch Linux.

### DIFF
--- a/src/HIE/Bios/Cradle.hs
+++ b/src/HIE/Bios/Cradle.hs
@@ -173,6 +173,11 @@ readCradleConfig yamlHie = do
 configFileName :: FilePath
 configFileName = "hie.yaml"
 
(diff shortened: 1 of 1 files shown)


fendor (approved):
LGTM, thanks!

fendor:
Acutally, I think it would be a good idea to add documentation about the ``-dynamic` flags to the cradle types. E.g. `we need this when ghc is dynamically linked` or something

fendor (approved):
LGTM, thanks!

PR #288, opened 2021-03-09
Fix tests broken by #287

berberman:


jneira (approved):
lgtm, thanks for the quick fix

PR #291, opened 2021-04-14
Slight refactoring and more doc-comments

masaeedu:
I was reading through the codebase to try and understand what it does and encountered a bit that was a bit confusing to read. I tried breaking it apart a bit and added some doc comments, hopefully you agree that this represents a slight improvement.

I also encountered a type that doesn't seem to be used by any of the functions that ask for it (they use an `_` prefixed variable to discard it). Can this type just be removed?

### DIFF
--- a/src/HIE/Bios/Cradle.hs
+++ b/src/HIE/Bios/Cradle.hs
@@ -27,6 +27,7 @@ import Control.Exception (handleJust)
 import qualified Data.Yaml as Yaml
 import Data.Void
 import Data.Char (isSpace)
+import Data.Bifunctor (first)
 import System.Process
 import System.Exit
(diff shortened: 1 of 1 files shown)


jneira:
Hi, thanks for the pr, the comment and refactor of implicitCongi looks good.
Not sure about removing CradleOpts though, maybe @fendor could help us to understand why it exists.

masaeedu:
@jneira In the interests of making this easier to review I've removed the CradleOpts stuff and will put that in a separate PR.

jneira (approved):
Thanks for separate it in another pr, lgtm

fendor (commented):
Nice, thank you!

PR #292, opened 2021-04-14
Use Data.Monoid.Last from base

masaeedu:
It seems like the `Maybe ∘ Data.Semigroup.Last` business we're doing is effectively just giving us `Data.Monoid.Last`, which we can import directly.


fendor (approved):
Awesome, glad to see this code simplified

PR #293, opened 2021-04-14
Remove CradleOpt type

masaeedu:
I was reading through the code to try and understand it better and encountered this type that doesn't seem to be used by any of the functions that ask for it (they use an _ prefixed variable to discard it).

Can this type just be removed?

### DIFF
--- a/hie-bios.cabal
+++ b/hie-bios.cabal
@@ -1,6 +1,6 @@
 Cabal-Version:          2.2
 Name:                   hie-bios
-Version:                0.7.6
+Version:                0.8.0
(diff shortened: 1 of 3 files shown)


masaeedu:
cc @jneira @fendor

fendor (approved):
Seems about right, nothing is using this, and afaict this is a relic from hie-bios' foundling days

fendor:
As this is a public API change, it requires a major version bump.

@masaeedu Would you mind updating `hie-bios.cabal` as well?

fendor (approved):

PR #294, opened 2021-04-25
Don't look for NIX_GHC_LIBDIR as it is redundant

### LINKED ISSUE #273: Using NIX_GHC_LIBDIR is surprising and undocumented
It is surprising (and undocumented) to check for `NIX_GHC_LIBDIR` in order to work out which libdir to use.

My situation was that I explicitly set the compiler for a bios cradle but NIX_GHC_LIBDIR was still being incorrectly consulted to pick the libdir of a different compiler.

The workaround was to call `unset NIX_GHC_LIBDIR` in the bios script.

fendor:
Closes #273


mpickering:
LGTM.

bgamari:
Yes, I agree, there is not reason to look at `NIX_GHC_LIBDIR`.

jneira:
it is being used in ghcide, so we should remove it there too
https://github.com/haskell/haskell-language-server/blob/17fce672402f22944527c616a0c4015df1ffb750/ghcide/session-loader/Development/IDE/Session/VersionCheck.hs#L17

PR #300, opened 2021-07-06
Add compatbility for GHC 9.0 and 9.2

fendor:
Avoids dependency on `ghc-api-compat` so that `hie-bios` can actually be uploaded to Hackage.

Additionally, moves `parseTargetFiles` to `Gap.hs` since GHC 9.2 exposes the function we want :confetti_ball:.

supersedes #285


jneira:
It would be great make haskell gha download the ghc-9.2 prerelase to test it in ci (maybe it is possible? ghcup can install the ghc prerelase and gha uses ghcup under the hood)

fendor:
Related: https://github.com/haskell/actions/issues/55

I would like to avoid implementing custom support, so I would wait for this feature to be implemented.

jneira (approved):
LGTM, thanks for this double support

src/HIE/Bios/Ghc/Gap.hs
```diff
@@ -131,25 +291,234 @@ outType :: MatchGroup GhcTc LExpression -> Type
...
+addWay' :: Way -> DynFlags -> DynFlags
+addWay' w dflags0 =
+   let platform = targetPlatform dflags0
+       dflags1 = dflags0 { targetWays_ = Platform.addWay w (targetWays_ dflags0) }
+       dflags2 = foldr setGeneralFlag' dflags1
+                       (Platform.wayGeneralFlags platform w)
+       dflags3 = foldr unSetGeneralFlag' dflags2
+                       (Platform.wayUnsetGeneralFlags platform w)
+   in dflags3
+#endif
```
berberman, line 426: IIRC, we don't need this in later GHC versions, as we can use `GHC.Driver.Session.setDynamicNow` directly.

Issue #269, opened 2020-12-09
Proposal: better multi component support


pepeiborra:
The ghcide README says:

> There is a known issue where if you have three components, such that A depends on B which depends on C then if you load A and C into the session but not B then under certain situations you can get strange errors about a type coming from two different places. See this repo for a simple reproduction of the bug.

This is because of the way multicomponent support works in ghcide. Loading A and C creates a fake package AC which results from the union of the flags and targets in A and C. So AC depends on B because A depends on B, and transitively on C because B depends on C. As a result the types in C are loaded twice, once for AC and once for B, and even if they have the same names they are not the same types, which leads to the weird errors.

To fix this, we need to load B in the session, obtaining ABC. The user can do this by opening any B file in their editor, but ideally this would be done automatically by ghcide. Unfortunately, ghcide has no way of knowing about this! 

My proposal is to extend `getComponentOptions` to accept the set of components already loaded and produce a set of component options:

```
getCompilerOptions :: FilePath -> [LoadedComponent] -> Cradle a -> IO (CradleLoadResult [ComponentOptions])
```

This would change the example in the ghcide README as follows:

1. The user opens A.hs from A in the editor,
2. ghcide asks the cradle about A.hs *letting it know that the current set of components is []*
3. hie-bios returns the component options for the set [A]
3. ghcide creates and loads the component A
4. The user opens C.hs from C in the editor
5. ghcide asks the cradle about C.hs letting it know that the current set of components is [A]
6. hie-bios returns the component options for the set [B,C]
7. ghcide creates and loads the component ABC

To implement the new `getCompilerOptions`:
* the Stack and Cabal cradles will need to interrogate the build system to find out component dependencies
* multi cradles will need to interrogate all the single cradles, which will probably be ridiculously expensive
* Bios cradles will need to extend the protocol with the program to pass the information of the loaded components

This proposal is very much up for discussion, and note that I have not prototyped any of this so I have no idea of whether it will work.

fendor:
@OliverMadine I think this issue summarises the idea succinctly. 

Changing hie-bios' API is relatively trivial, no real deal here.

The main challenge is to extract the required information from cabal and stack. The information we need is basically, given a source file, we need to know how we can compile the given file (ideally the whole component the file belongs to). For multi-component support, we need the information for all components at the same time, e.g. what are the components and what GHC options do we need to compile them.

[...]

Issue #255, opened 2020-10-03
Multi-component REPL


robrix:
I have configured a bunch of different projects to use a script to load multiple components in `ghcide`. This is no longer necessary as the multi-component support in `hie-bios` is working beautifully! 💖 However, I also use these scripts to provide a multi-component REPL—to load `ghci` with the right search paths, warnings, etc. for the library, tests, and executables.

Unfortunately, maintaining these `ghci-flags` scripts is kind of a drag; I’d much rather configure `hie-bios` for multiple components (which I will do for editor integration anyway) and then ask it. However, `hie-bios flags` expects to be called with a source file, but I wouldn’t know that until I’m actually running in `ghci`, and even then, it’s not really how `ghci` works; it doesn’t use a different set of options per file.

I’m definitely ok with compromising there a little bit by e.g. swapping out the `-outputdir`—it’s what my scripts to return the flags do anyway—but it would be super convenient to get the configuration for each component from `hie-bios` with a single call rather than having to find or nominate a source file within each component and ask `hie-bios` for the flags for each one by one. Maybe the interface could be `hie-bios flags` without passing paths.

Is this something you’d consider supporting?

fendor:
Hi, sorry for the very delayed response.

Can you elaborate what changes you would require? Changes to the cli are not a problem if it would help your use-case (no idea whether this is still relevant to you)

robrix:
Still very much relevant 😊 I think a `hie-bios flags` command would be excellent, or alternatively a `hie-bios repl` command  which launches `ghci` with the union of the flags for the various components `hie-bios` has been configured for. To be honest, I’d prefer the former, as it potentially serves other use cases as well, and gives me more latitude in passing extra flags to `ghci` to e.g. configure the RTS or load scripts. It also seems a little more in keeping with `hie-bios`’ design thus far.

So in brief, I’d aim for:

1. an `hie-bios flags` (or w/e name is preferred) command
2. which emits flags suitable for passing to `ghci`, one per line,
3. including paths for interface files, object files, sources, autogen’d files, headers, packages, warning flags, and so on,
4. for all `cabal` package components for which `hie-bios` has been configured (and therefore unioning multiple components’ flags, compromising as necessary to deal with mutually exclusive/contradictory flags), and
5. without requiring the caller to specify the paths to any individual source files like the usual `hie-bios` calls require.

For comparison, the scripts I’m using now are mostly copy-pasta after some initial configuration quite some time ago. Here’s an example:

[...]

PR #301, opened 2021-07-09
WIP for cabal show-build-info

fendor:
Use features from https://github.com/haskell/cabal/pull/7478

Starts work on #269

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -1,1 +1,6 @@
 packages: .
+
+source-repository-package
+    type: git
+    location: https://github.com/fendor/cabal-build-info
+    tag: 80cf21bdde938255e5736c4a35a74fecd9005029
--- a/exe/Main.hs
+++ b/exe/Main.hs
@@ -14,6 +14,7 @@ import HIE.Bios.Ghc.Check
 import HIE.Bios.Ghc.Gap as Gap
 import HIE.Bios.Internal.Debug
 import Paths_hie_bios
+import qualified Data.List.NonEmpty as NE
 
 ----------------------------------------------------------------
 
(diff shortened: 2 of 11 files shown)


georgefst:
I'm curious, with upstream support in Cabal seemingly now implemented, is there anything blocking this? There are a lot of issues in HLS which mention this as the solution.

fendor:
Yeah, there are unfortunately a couple of open issues :sweat_smile: Also, #255 essentially implements proper multi-home unit support without `show-build-info`.

georgefst:
But #409 doesn't solve the issue of the user having to manually load all the right components to prevent errors like https://github.com/haskell/haskell-language-server/issues/1370. I thought the point of `show-build-info` was that it finally gives us the information we need in order to know which components to load.

georgefst:
> But #409 doesn't solve the issue of the user having to manually load all the right components to prevent errors like [haskell/haskell-language-server#1370](https://github.com/haskell/haskell-language-server/issues/1370).

Or maybe it does? I just noticed from https://github.com/haskell/haskell-language-server/issues/3738 that the latest HLS release doesn't incorporate those commits yet.

fendor:
> But https://github.com/haskell/hie-bios/pull/409 doesn't solve the issue of the user having to manually load all the right components to prevent errors like https://github.com/haskell/haskell-language-server/issues/1370. I thought the point of show-build-info was that it finally gives us the information we need in order to know which components to load.

It does! Conceptually, `cabal repl --enable-multi-repl` computes all the components you need to load into ghci, invokes ghci via `ghci @args.unitA.txt @args.unitB.txt` and HLS simply looks at these args and sets up the session appropriately.
Currently, the only advantage seems to be that `show-build-info` uses a well-defined interface instead of the hacky `cabal repl -w hie-bios-ghc` solution.

georgefst:
Aha, awesome! Well then, I very much look forward to HLS being released with the new hie-bios.

PR #304, opened 2021-08-23
Change API to emit NonEmpty ComponentOptions

fendor:
Introduces changes necessary for #301 

This PR can be merged before #301, and released, then HLS can be updated for the new API.


fendor:
cc @OliverMadine

jneira (approved):
horror vacui 👍

fendor:
What are the semantics of the provided `NonEmpty`? Do we  require that the component the file path is part of has been loaded and is the first element of the `NonEmpty`? How can we communicate that we can not load a component because one of our dependencies has failed to build?
I propose, mainly for the sake of argument, a new data type instead of `NonEmpty`:
```haskell
data LoadResult = 
  { loadComponent :: Maybe ComponentOptions,
  , componentDependencies :: Maybe (NonEmpty ComponentOptions)
  }
``` 

intended semantics: it is possible to extract component dependencies and load these immediately, but we were unable to load the component the file target is part of.
No guarantee for the order of dependencies. They may not exist, but if they do exist, there must exist at least. This type is isomorphic to just `[ComponentOptions]`, but maybe expresses the intent better.

Opinions?

jneira:
> Opinions?

i like the precision of the new data type and it can be extended if we find out other nuances

src/HIE/Bios/Types.hs
```diff
@@ -44,21 +44,60 @@ data ActionName a
...
+-- Sets the 'ComponentOptions' as 'loadResultComponent'.
+mkSimpleLoadResult :: ComponentOptions -> LoadResult
+mkSimpleLoadResult opts = LoadResult
+  { loadResultComponent = Just opts
+  , loadResultDependencies = []
+  }
+
+-- | Helper to access the main component if there is one.
+pattern Main :: a -> LoadResult' a
+pattern Main opts <- (loadResultComponent -> Just opts)
```
fendor, line 99: We can have more of these such as:
```
Deps :: [a] -> LoadResult' a
MainAndDeps :: a -> [a] -> LoadResult' a
```

for typesafe pattern matching.

jneira:
@fendor hi! any chance to get this merged to make progress in enable rename plugin in hls?

fendor:
It is right now still kind of a WIP. Do you have opinions on naming?

src/HIE/Bios/Types.hs
```diff
@@ -44,21 +44,60 @@ data ActionName a
...
+data LoadResult' a = LoadResult
+  { loadResultComponent :: Maybe a
+  -- ^ Component options for the FilePath that produced this 'LoadResult'.
+  -- See 'CradleAction.runCradle' for information on how to produce a 'LoadResult'.
+  --
+  -- This field can be 'Nothing' to indicate that loading partially failed/succeeded.
+  , loadResultDependencies :: [a]
+  -- ^ Direct or indirect dependencies from the component from above.
+  -- Indirect means that it is not required that 'ComponentOptions' in this list
+  -- are required dependencies of 'loadResultComponent'. It is specifically allowed
```
fendor, line 78: The sentence makes no sense, what I mean is, that items here are either local dependencies in some form, or completely independent but can be loaded, too.
It might still be important to have some kind of order on them.

jneira:
> It is right now still kind of a WIP. Do you have opinions on naming?

about https://github.com/haskell/hie-bios/pull/304#issuecomment-919217913? i like the `LoadResult` data type

src/HIE/Bios/Types.hs
```diff
@@ -44,21 +44,60 @@ data ActionName a
...
   deriving (Functor)
 
 instance Show a => Show (CradleAction a) where
   show CradleAction { actionName = name } = "CradleAction: " ++ show name
 
+type LoadResult = LoadResult' ComponentOptions
+
+-- | Record for expressing successful loading.
+-- Can express partial success.
+data LoadResult' a = LoadResult
```
jneira, line 69: What about somethin like

```haskell
data ComponentInfo = ComponentInfo
  { componentOptions :: Maybe a
   componentDependencies: [a]
  }
```

It would be generic enough to allow add more component information afterwards

fendor: Hm, maybe we can drop the prefix somehow to be less "Component"-y? However, since the record field prefix will probably still be required, probably pointless.

jneira: yeah, `Info` sounds good too

michaelpj:
What's the status of this?

fendor:
It is on hold because it is semi-blocked on https://github.com/haskell/cabal/pull/7500 to get reliably the build-info location.

Technically, it is not blocked, the API could be changed without cabal support, but then we have a slightly harder time to test the implementation. Also, if there is no cabal or stack support for it, I see barely an incentive to implement it, currently.

fendor:
Superseded by #409

PR #308, opened 2021-08-30
Prepare Release 0.7.6

fendor:
This PR is not intended for merge, it backports backwards compatible changes to `0.7.5` to avoid a major version bump.

When CI is green, the PR will be pushed into the main repository, tagged and released to Hackage.


fendor:
Released to Hackage https://hackage.haskell.org/package/hie-bios-0.7.6

cc @jneira

PR #312, opened 2021-09-20
hie-bios --help now works even with a faulty hie.yaml

teto:
Running `hie-bios --help` in a folder with a faulty hie.yaml would crash
straightaway with for instance:
```
hie-bios: AesonException "Error in $.cradle.cradle: Not a known cradle type. Possible are: cabal, stack, bios, direct, default, none, multi"
```
without displaying the help. This should not happen.

NB: Would you mind if I opened a PR to add a flake.nix ? like haskell-language-server does.

### DIFF
--- a/exe/Main.hs
+++ b/exe/Main.hs
@@ -59,13 +59,13 @@ main :: IO ()
 main = do
     hSetEncoding stdout utf8
     cwd <- getCurrentDirectory
+    cmd <- execParser progInfo
     cradle <-
         -- find cradle does a takeDirectory on the argument, so make it into a file
         findCradle (cwd </> "File.hs") >>= \case
           Just yaml -> loadCradle yaml
           Nothing -> loadImplicitCradle (cwd </> "File.hs")
 
-    cmd <- execParser progInfo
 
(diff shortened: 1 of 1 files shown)


fendor:
LGTM, thanks!

fendor (approved):

fendor:
> NB: Would you mind if I opened a PR to add a flake.nix ? like haskell-language-server does.

What would be the guarantees (e.g. you can build the executable, etc...)?
I am not generally opposed to it, but even as a nix user, I often feel the cost of supporting builds with nix outweigh the benefits. We need CI, an experimental nix feature, and I worry nix pushes away new contributors.

teto:
I wanted to provide just enough to run `nix develop`, I dont want to commit to install a CI etc. I usually use nix just to get some system libraries + ghc since it is indeed hard to build haskell stuff via nix (I gave haskell.nix a try but it has its own issues)

fendor:
If we don't commit to maintain it, I personally see little value in it. You can open a PR so others can see and use it, but I don't think we would merge it then

PR #317, opened 2021-11-25
Remove hie.yaml

fendor:
Remove hie.yaml as it is not necessary for HLS anymore
Remove hie.yaml.back which was an accident


jneira (approved):
lgtm, good clean up

PR #322, opened 2021-12-28
Fix compilation with GHC 9.0.2

sternenseemann:
GHC commit [7f1fff5][1] which is included in 9.0.2 is a backport of an
API change to the IncludeSpecs record of 9.2.1 which is already
accounted for by an `__GLASGOW_HASKELL__` CPP guard. For 9.0.2
compatibility, we simply need to hit this guard for 9.0.2 as well, but
not for 9.0.1.

[1]: https://gitlab.haskell.org/ghc/ghc/-/commit/7f1fff501faa9525ca09bde076059a65388d638a

### DIFF
--- a/src/HIE/Bios/Ghc/Gap.hs
+++ b/src/HIE/Bios/Ghc/Gap.hs
@@ -260,7 +260,7 @@ mapOverIncludePaths f df = df
       G.IncludeSpecs
           (map f $ G.includePathsQuote  (includePaths df))
           (map f $ G.includePathsGlobal (includePaths df))
-#if __GLASGOW_HASKELL__ >= 902
+#if MIN_VERSION_GLASGOW_HASKELL(9,0,2,0)
           (map f $ G.includePathsQuoteImplicit (includePaths df))
 #endif
 #else


fendor (approved):
LGTM, waiting for CI to finish

sternenseemann:
I could add 9.0.2 to the CI matrix in this PR as well if you want?

fendor:
Sure, is that already possible? E.g. does haskell-actions have ghc 9.0.2 support?

sternenseemann:
I don't know, I just assumed it would throw the version against ghcup, but maybe it doesn't?

fendor:
It does, but at least for me ghcup does not have support for 9.0.2 yet.

sternenseemann:
Well just take my word then that it works with 9.0.2 :)

fendor:
Gladly :D Do you require for nixpkgs a release to get this in?

sternenseemann:
We're using 0.7.6 anyways atm (due to stackage probaby?), so I think we need to apply it as a patch either way which should be fine (unless you fancy releasing a 0.7.7 with a backported patch). 0.8.0 is unused atm, so an 0.8.1 release is not urgent.

jneira:
> It does, but at least for me, ghcup does not have support for 9.0.2 yet.

but chocolatey does, so ci should be succesful for windows :-P

PR #323, opened 2021-12-28
Drop hspec-expectations asserts as they don't get rendered properly by tasty

fendor:
I introduced `hspec-expectations` because I prefer the asserts over the infix operators from `tasty-hunit` and the libraries feels more batteries-included.

Tasty doesn't render errors from those assert properly, which makes it way harder to read once a test fails. Thus, drop them.


jneira (approved):
lgtm

PR #324, opened 2022-01-06
Add CI support for GHC 9.0.2 and GHC 9.2.1

fendor:
Skips Windows CI as one of the dependencies (unix-compat) doesn't build for GHC 9.0.2 and 9.2.1

Once green, will be released.

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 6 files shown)


jneira:
> Skips Windows CI as one of the dependencies (unix-compat) doesn't build for GHC 9.0.2 and 9.2.1
> 
> Once green, will be released.

windows support could be kept with a remote repo source (see ghc-9.2 branch in his) in the cabal.project

fendor:
> windows support could be kept with a remote repo source

True, good idea to have it at least in the CI

Issue #315, opened 2021-10-31
Avoid CPP in Config.hs


fendor:
In https://github.com/haskell/lsp/pull/360, compatibility with `aeson >= 2` is achieved without using CPP, we want that too, because no one likes CPP!

fendor:
Rewriting *should* be possible, however, the parsers of `Config.hs` are hard to follow, so this might be an opportunity to rewrite the parsers.

ThomasCrevoisier:
:wave: I can try to give a shot at this :) Though I would need some pointers before doing so.

From what I see, the only impacted points are parsing HashMap and the use of `KeyMap` in `parseSingleOrMultiple`, right ?

> however, the parsers of Config.hs are hard to follow, so this might be an opportunity to rewrite the parsers

Do you have something specific in mind ?

fendor:
Hello! Happy to hear it!

True, I think so as well, these are the only locations where we access the Keys.
To be frank, I am not fully sure right now, whether we can avoid all CPP, but any reduction would be great!

> Do you have something specific in mind ?

I feel like we have a number of code smells, for example, we dispatch manually the cradle type in https://github.com/haskell/hie-bios/blob/master/src/HIE/Bios/Config.hs#L158.

Then the individual parsers feel a little off as well, take https://github.com/haskell/hie-bios/blob/master/src/HIE/Bios/Config.hs#L203 where we explicitly check the size of the Aeson object to determine what values are possible and provide a proper error message for the user (e.g. when we fail to parse a `hie.yaml`). Maybe we can improve this parser? Or can at least generate the same error message without hard-coding the available values?

In general, everything you would consider a code-smell can and should be re-worked!

The structure and code of the types `CradleConfig` and `CradleType` can be changed if you think it improves the code-quality.

If you are not sure about something, you can either ask here or ping me in irc/matrix (libera.chat #haskell-language-server, I have the same name)

fendor:
Closing, has been fixed by #329

PR #329, opened 2022-02-16
Refactor parsing of hie.yaml files

ThomasCrevoisier:
Takes care of #315 

This introduces a `Hie.Bios.Config.YAML` module which defines datatypes a bit closer to the YAML structure.
It allows writing simpler FromJSON instances (without using CPP !) and should be equivalent to the types existing in the `Hie.Bios.Config` module.


fendor:
Hi! Looking good!

Since CI has a failure, can you add the failing hie.yaml to the parser-tests? If I read this correctly, should be this one: https://github.com/haskell/hie-bios/blob/master/tests/projects/failing-cabal/hie.yaml

ThomasCrevoisier:
@fendor all should be good now !

fendor (changes requested):

fendor:
Unfortunately, the generic variant suffers currently from the issue of ignoring unknown keys. Example:
```yaml
cradle:
  cabal:
    comp:
    - "exe:hie-bios"
```

is an invalid `hie.yaml`. 

and currently hie-bios reports:
```
hie-bios: AesonException "Error in $.cradle: Not a valid Cabal configuration, following keys are allowed: component"
```
but with this patch, this becomes:
``` 
Config {cradle = CradleConfig {cradleDependencies = [], cradleType = Cabal {component = Nothing}}}
``` 

It successfully parses, and ignores that you mistyped or misnamed the argument.

Can we do better than that?

ThomasCrevoisier:
Oh I see. I've got few questions on this point.

As I've seen in the test suite, empty objects are allowed right ?

Regarding unknown keys, what should be the behavior when the object we provide contains a valid config but with extra keys that we will ignore ?

Failing on missing keys is pretty straightforward, but failing on extra unwanted keys might be a bit trickier to do.

fendor:
> Regarding unknown keys, what should be the behavior when the object we provide contains a valid config but with extra keys that we will ignore ?

Until now, we threw an error message. The point being that you notice you didn't write the correct configuration and something is not right.
The idea being that subtle typos are discovered ASAP, and not just ignored.

At the moment, I think it is better if we throw an error if there are extra keys we don't know.

It is indeed trickier :/

ThomasCrevoisier:
f2f16ecf9d80d2d56b29f821216b7c607c3ecb10 is a proposal for fixing this particular point.

The downside is that we have to duplicate a bit the keys between parsing and checking but the implementation is relatively simple.

Let me know what you think !

fendor:
BTW, I did not forget, just didn't have time for an in-depth review (since this is important to get right). So far it looks good, and I am going to review in depth until the end of the weekend.

ThomasCrevoisier:
Sure, absolutely no worries ! I won't be responsive this week-end, but I'll remain available to work on it next week :)

fendor:
I see your point and did not consider this angle before. @ThomasCrevoisier Would it be hard to turn these errors into warnings?

fendor (approved):
As I don't want to keep you waiting, we are going to merge this as it is an improvement of the status-quo. We will create an issue for the remaining issue

ThomasCrevoisier:
👋 Truely sorry @fendor, those past months have been quite busy and I forgot to come back to this PR 😞 

Feel free to move forward with it (I don't have any simple solution in mind regarding extra keys handling unfortunately), and again, sorry for the lack of reactivity.

fendor:
No need to apologise, life can be stressful (speaking from experience), and we all deeply appreciate the time you have spent on this!

PR #332, opened 2022-02-23
Put wrapper sources in tmp on windows

fendor:
Small behavioural regression from #282


fendor:
As pointed out by @jneira I misread the code from here: https://github.com/haskell/hie-bios/blob/0213509d32f394ebe1c8a56a3b37597a45b151c1/src/HIE/Bios/Cradle.hs#L503

PR #342, opened 2022-04-06
fix typos

askeblad:
src/HIE/Bios/Ghc/Load.hs

<hr>

src/HIE/Bios/Ghc/{Api.hs,Load.hs}

FWIW, Messager/messager is an obsolete form of Messenger/messenger.


askeblad:
@fendor evidently requires review

fendor:
Hi, thank you for fixing those typos!

> FWIW, Messager/messager is an obsolete form of Messenger/messenger.

It is taken from GHC which uses this form, so for consistency, we keep the `Messager/messager` names. (source: https://gitlab.haskell.org/ghc/ghc/-/blob/master/compiler/GHC/Driver/Main.hs#L738)

fendor (approved):

PR #347, opened 2022-05-05
Move logging from hslogger to co-log

michaelpj:
This lines up with the way that logging is being done in
`haskell-language-server` and `lsp`.

At the moment `hie-bios` is the only reason HLS has a `hslogger`
dependency, and it means that the logs from `hie-bios` are inconsistent
with those from the rest of the subsystems.


michaelpj:
This is what the output from HLS looks like after incorporating this change. You can see that the hie-bios logs now get formatted like everything else.

Before: 
image

After:
image

michaelpj:
cc @eddiemundo

src/HIE/Bios/Ghc/Load.hs
```diff
@@ -23,7 +26,23 @@ import qualified HscMain as G
...
+data Log =
+  LogLoaded FilePath FilePath
+  | LogTypechecked [TypecheckedModule]
+  | LogInitPlugins Int [ModuleName]
+  | LogSetTargets [(FilePath, FilePath)]
+  | LogModGraph ModuleGraph
+
+instance Pretty Log where
+  pretty (LogLoaded fp1 fp2) = "Loaded" <+> viaShow fp1 <+> "-" <+> viaShow fp2
+  pretty (LogTypechecked tcs) = "Typechecked modules for:" <> (cat $ map (viaShow . get_fp) tcs)
```
eddiemundo, line 39:
```suggestion
  pretty (LogTypechecked tcs) = "Typechecked modules for:" <+> (cat $ map (viaShow . get_fp) tcs)
```

src/HIE/Bios/Ghc/Load.hs
```diff
@@ -23,7 +26,23 @@ import qualified HscMain as G
...
+  pretty (LogInitPlugins n ns) = "Loaded" <+> viaShow n <+> "plugins, specified" <+> viaShow (length ns)
+  pretty (LogSetTargets ts) = "Set targets:" <+> viaShow ts
+  pretty (LogModGraph mod_graph) = "ModGraph:" <> viaShow (map ms_location $ Gap.mgModSummaries mod_graph)
```
eddiemundo, line 42:
```suggestion
  pretty (LogModGraph mod_graph) = "ModGraph:" <+> viaShow (map ms_location $ Gap.mgModSummaries mod_graph)
```

src/HIE/Bios/Ghc/Check.hs
```diff
@@ -14,37 +16,49 @@ import qualified DynFlags as G
...
 
+data Log =
+  LoadLog Load.Log
+  | LogAny T.Log
+  | forall a . Show a => LogCradle (Cradle a)
+
+instance Pretty Log where
+  pretty (LoadLog l) = pretty l
+  pretty (LogAny l) = pretty l
+  pretty (LogCradle c) = "Cradle: " <> viaShow c
```
eddiemundo, line 41:
```suggestion
  pretty (LogCradle c) = "Cradle:" <+> viaShow c
```

eddiemundo (changes requested):
LGTM. Only have super minor formatting tweaks. Seems like the tests don't compile though, and 8.6.5 needs an extension for the `forall`. Thanks for all the work, haven't had time to do anything except follow HLS superficially...

fendor:
Thank you for bringing co-log to hie-bios!

exe/Main.hs
```diff
@@ -1,9 +1,12 @@
 {-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 
 module Main where
 
 import Control.Monad ( forM )
+import qualified Colog.Core as L
 import Data.Version (showVersion)
+import Data.Text.Prettyprint.Doc
```
fendor, line 9: GitHub Actions scream a bunch of deprecation warnings here. Can we do something about that? I remember, there was an issue with my PR https://github.com/haskell/haskell-language-server/pull/2352

src/HIE/Bios/Cradle.hs
```diff
@@ -999,7 +1013,7 @@ type OutputName = String
 -- * The process is executed in the given directory.
 readProcessWithOutputs
   :: Outputs  -- ^ Names of the outputs produced by this process
-  -> LoggingFunction -- ^ Output of the process is streamed into this function.
+  -> LogAction IO (WithSeverity Log)
```
fendor, line 1016: I feel like the comment is important in this case, can we restore that?

fendor (commented):
This looks great, looking forward to it!

If you have troubles with the tests (because they are hard-to-read), give me a shout and I can help out!

exe/Main.hs
michaelpj: Yeah, I'm not sure what the best approach is. Pepe really wants to keep compatibility with old prettyprinter versions. That gives us two options: 1) use the deprecated import, 2) use CPP. I hate 1 less but I'm happy to take the other bad option also.

michaelpj:
Fixed the tests. Somehow I just didn't realise they existed :sweat_smile: I generally stopped them from issuing a step *every* log line, which seemed excessive, and instead just put in a couple of explicit steps.

CI is green.

exe/Main.hs
fendor: Eh, for now, just leave it be. Eventually we can just drop it, probably.

fendor (approved):
LGTM, awesome!

PR #348, opened 2022-05-06
Demote process output to Debug severity

michaelpj:
This is perhaps a more controversial change. The process output is quite
noisy, and arguably not of interest unless you're debugging an issue.

The counterargument is that if something *does* go wrong with running the
cradle command, it's useful to see it. Perhaps we could collect the
output and only emit it on failure?

### DIFF
--- a/src/HIE/Bios/Cradle.hs
+++ b/src/HIE/Bios/Cradle.hs
@@ -1027,7 +1027,7 @@ readProcessWithOutputs outputNames l workDir cp = flip runContT return $ do
 
     -- Windows line endings are not converted so you have to filter out `'r` characters
   let loggingConduit = C.decodeUtf8  C..| C.lines C..| C.filterE (/= '\r')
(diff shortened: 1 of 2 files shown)


fendor:
IIRC, aren't these logs used for displaying progress messages in VSCode? (Maybe I am mistaken)

michaelpj:
Uh. I have no idea. That would be... surprising.

fendor:
Hm, maybe not. Maybe we should add a different Constructor in this case? Such as `CradleLoadingProcess`? 
The idea being that we can actually use these logs to show something to the user.

michaelpj:
Okay, I checked. As far as I can tell, at the moment these go via `hslogger`, and the `hslogger` output is _just_ logged, and not used for anything else. So I think we're safe.

> Maybe we should add a different Constructor in this case? Such as CradleLoadingProcess?

Yeah, good idea.

michaelpj:
Added a special constructor for it.

fendor (approved):

PR #353, opened 2022-06-16
Cabal cradle: change error message on failure

smatting:
This changes the first line of the error message that is shown when the cabal cradle fails to

```
Failed to run ["cabal", "v2-repl",  ..] in directory  "/home/foo/...". See below for full command and error.
```

(the error message was `Failed to parse result of calling cabal` before, which is less informative).

cc @fendor

### DIFF
--- a/src/HIE/Bios/Cradle.hs
+++ b/src/HIE/Bios/Cradle.hs
@@ -699,13 +699,31 @@ cabalAction
   -> FilePath
   -> CradleLoadResultT IO ComponentOptions
 cabalAction workDir mc l fp = do
-  cabalProc <- cabalProcess workDir "v2-repl" [fromMaybe (fixTargetPath fp) mc] `modCradleError` \err -> do
+  let
+    cabalCommand = "v2-repl"
+    cabalArgs = [fromMaybe (fixTargetPath fp) mc]
+
+  cabalProc <- cabalProcess workDir cabalCommand cabalArgs `modCradleError` \err -> do
       deps <- cabalCradleDependencies workDir workDir
       pure $ err { cradleErrorDependencies = cradleErrorDependencies err ++ deps }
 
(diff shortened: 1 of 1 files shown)


src/HIE/Bios/Cradle.hs
```diff
@@ -699,27 +699,39 @@ cabalAction
...
+        , unlines output
+        , unlines stde
+        , unlines $ args
+        , "Process Environment:"]
+        <> prettyProcessEnv cabalProc
+
+  when (ex /= ExitSuccess) $ do
+    deps <- liftIO $ cabalCradleDependencies workDir workDir
+    let cmd = show (["cabal", cabalCommand] <> cabalArgs)
+    let errorMsg = "Failed to run " <> cmd <> " in directory \"" <> workDir <> "\". See below for full command and error."
```
fendor, line 724: I am not sure that "See below for full command and error." is going to be much helpful to users, since where is "below"?

Not sure this sentence is necessary.

fendor (approved):
LGTM, one small detail up for disucssion, imo

src/HIE/Bios/Cradle.hs
smatting: Maybe "See stderr for full command and error"?. I think it's helpful to point out that there's more information available

fendor: I agree, but if this is shown in the LSP dialog pop-up, "below" has no meaning. However, I was actually unable to provoke that pop-up window, is this only shown as a diagnostic?

michaelpj: I think you would need to build HLS with a `hie-bios` that includes https://github.com/haskell/hie-bios/pull/347 to get errors automatically shown to the user. I do agree that "below" is ambiguous here. Maybe "consult the logs"?

smatting: @fendor Yes, it's only a diagnostic (on othe first line of the module). @michaelpj :+1: I changed the error msg like you suggested

fendor (approved):
LGTM, thanks!

PR #357, opened 2022-08-18
Add support for non-standard cabal.project file locations to cabal cradles

### LINKED ISSUE #263: Add `project-file` config option to cabal cradle
* Like we added `stack-yaml` for the stack cradle
* It could be used to workaround for #195 (and other issues?)
* And the api would be symmetric between both tools

fendor:
Add parser tests for non-standard `cabal.project` file locations
Add bios tests for projects with non-standard `cabal.project` file locations
Follows the same structure as `stackYaml` feature for stack cradles.

Fixes #263

### DIFF
--- a/src/HIE/Bios/Config.hs
+++ b/src/HIE/Bios/Config.hs
@@ -10,6 +10,7 @@ module HIE.Bios.Config(
     CabalType,
     pattern CabalType,
     cabalComponent,
+    cabalProjectFile,
     StackType,
     pattern StackType,
     stackComponent,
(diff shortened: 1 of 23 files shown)


src/HIE/Bios/Config.hs
```diff
@@ -56,17 +57,17 @@ data Callable = Program FilePath | Command String
     deriving (Show, Eq)
 
 data CabalType
```
michaelpj, line 59: Good time to add some Haddock for these types? :innocent:

src/HIE/Bios/Cradle.hs
```diff
@@ -455,22 +457,32 @@ callableToProcess (Program path) file = do
...
 ------------------------------------------------------------------------
+
+projectFileProcessArgs :: CradleProjectConfig -> [String]
+projectFileProcessArgs (ExplicitConfig prjFile) = ["--project-file", prjFile]
+projectFileProcessArgs NoExplicitConfig = []
+
+projectLocationOrDefault :: CradleProjectConfig -> [FilePath]
+projectLocationOrDefault = \case
+  NoExplicitConfig -> ["cabal.project", "cabal.project.local"]
+  (ExplicitConfig prjFile) -> [prjFile, prjFile <.> "local"]
```
michaelpj, line 468: Should the explicit config include the .local file? Or will cabal pick it up automatically if it has the main project file?

src/HIE/Bios/Cradle.hs
```diff
@@ -783,31 +810,42 @@ cabalWorkDir wdir =
...
-data StackYaml
-  = NoExplicitYaml
-  | ExplicitYaml FilePath
+data CradleProjectConfig
+  = NoExplicitConfig
+  | ExplicitConfig FilePath
+
+instance Semigroup CradleProjectConfig where
+  NoExplicitConfig <> s = s
+  s@(ExplicitConfig _) <> _ = s
```
michaelpj, line 825: suspicious. Should we just use `First CradleProjectConfig`?

michaelpj (approved):
Seems nice!

src/HIE/Bios/Cradle.hs
fendor: These instances aren't used anyway, maybe we should just remove them

src/HIE/Bios/Cradle.hs
fendor: This is used for cabal cradle dependencies, e.g. telling HLS that the cradle needs to be reloaded if either of the two changes. IIRC, that behaviour is in line with cabal.
However, the name is misleading.

Kleidukos:
@fendor what's the status on this PR? Do you think you can spare some time to solve the conflicts / rebase it?

Kleidukos:
@fendor I personally don't, but a post on the Discourse could certainly gauge the interest for this feature :)

michaelpj (approved):

PR #365, opened 2022-08-30
Actually run the bios-tests when tool-deps are ignored

fendor:
Apparently I messed this up since https://github.com/haskell/hie-bios/pull/330, which means CI didn't run any bios tests since about six months.

So we either can applaud ourselves for testing locally, or some releases are borked :)


hasufell (approved):

hasufell:
Definitely applause.

fendor:
Finally, let's get this merged.

PR #367, opened 2022-09-12
Add minimum bound on aeson

wz1000:
hie-bios 0.10.0.0 needs a revision on hackage

PR #368, opened 2022-09-12
Compatibility with aeson 1.5

wz1000:
Incompatible with https://github.com/haskell/hie-bios/pull/367

Should apply #367 to a branch and do the hackage revision

### DIFF
--- a/hie-bios.cabal
+++ b/hie-bios.cabal
@@ -150,7 +150,7 @@ Library
   autogen-modules:      Paths_hie_bios
   Build-Depends:
                         base                 >= 4.8 && < 5,
-                        aeson                >= 2.0 && < 2.2,
+                        aeson                >= 1.5 && < 2.2,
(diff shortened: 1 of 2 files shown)


hasufell:
hie-bios is not used by anyone else than HLS? I think supporting ancient dependency versions is not very important

fendor (approved):
LGTM, however, I would like to have some automatic way of checking that it builds with a certain version...

fendor:
> hie-bios is not used by anyone else than HLS? I think supporting ancient dependency versions is not very important

Issue is that not all HLS versions only build with aeson-2 I believe.

hasufell:
> > hie-bios is not used by anyone else than HLS? I think supporting ancient dependency versions is not very important
> 
> Issue is that not all HLS versions build with aeson-2.

What's the problem with locking those into an old hie-bios?

Or did you mean GHC versions?

fendor:
I meant ghc versions, yes.

PR #374, opened 2022-10-06
DRAFT: allow specifying --stack-work

### LINKED ISSUE #345: Allow overriding workdir in `hie.yaml`
### Context
At work we use stack with HLS. But we also use stack to run our application/tests. By default HLS will build things into the same stack workdir as regular calls to the stack command line. We've found this leads to excessive rebuilding as HLS and regular stack "fight".

### Feature request
`stack` cradles should have a `workdir` setting that allows overriding the workdir so that it doesn't interfere with non-HLS usages of stack.
I feel like `cabal` could have a similar setting but I don't know if this is a problem that cabal faces.

### Workaround
Right now what I do is copy our `stack.yaml` and append a line that specifies a custom work-dir and then tell hie-bios to use that modified stack.yaml file.

dten:
resolves #345

annoyingly even though this is entirely optional it's a breaking change for implicit-hie-cradle because it uses the sole pattern
https://github.com/Avi-D-coder/implicit-hie-cradle/blob/master/src/Hie/Implicit/Cradle.hs#L75

### DIFF
--- a/src/HIE/Bios/Config.hs
+++ b/src/HIE/Bios/Config.hs
@@ -14,6 +14,7 @@ module HIE.Bios.Config(
     pattern StackType,
     stackComponent,
     stackYaml,
+    stackStackWork,
     CradleType(..),
     Callable(..)
     ) where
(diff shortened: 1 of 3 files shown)


dten:
right now it seems to be missing something.. any guess?

fendor:
Hello, thank you for your contribution!

I agree that it needs to be specifiable to change the work-dir, I am just wondering right now, whether it would be better to have something more generic. I am thinking about allowing to specify environment variables, then you could have 
```yaml
cradle:
  stack:
environment:
  - STACK_WORK=.stack-work-directory
```

What do you think about that?

dten:
your amazing suggestion has kind of shown me that both changes feel a bit redundant right now. I found that the the code plugin and sublime plugins both allow specifying environments variables to add to the env of hls when running.

so for any additional stack arguments which can be passed via env this functionality is already easily usable _and_ lets you use the implicit cradle.

dten:
i'll close this for now, but if you think the environment option is still useful to anyway i might pick that up again

Issue #372, opened 2022-09-30
Log at INFO before running commands


michaelpj:
Especially potentially long-running ones like calling cabal or stack.

I demoted the process output to DEBUG, but that now means there's no feedback when we might be slowly building all the project deps. It would be helpful to emit an INFO message to tell the user what we're doing, e.g.
```
INFO: executing command: cabal repl ...
```
or even
```
INFO: invoking build tool to determine build flags (this may take some time depending on the cache)
INFO: executing command: cabal repl ...
```

PR #375, opened 2022-10-08
Add logs over commands

blackheaven:
I'm tackling #372 with a twist:

* Long running command is at `INFO`
* Commands per se are at `DEBUG` level

I think it gives enough feedback for day-to-day use, but allows to have a more verbose if needed.

Regarding design, it introduces a lot of burden, I wonder if I'd better to integrate it in `Cradle`.

WDYT?

### DIFF
--- a/exe/Main.hs
+++ b/exe/Main.hs
@@ -75,7 +75,7 @@ main = do
       logger = L.cmap printLog L.logStringStderr
 
     res <- case cmd of
-      Check targetFiles -> checkSyntax logger cradle targetFiles
+      Check targetFiles -> checkSyntax logger logger cradle targetFiles
       Debug files -> case files of
         [] -> debugInfo logger (cradleRootDir cradle) cradle
         fp -> debugInfo logger fp cradle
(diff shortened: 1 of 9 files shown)


fendor:
Sorry for the long delay! I just kept putting off the work to review :sweat_smile:

fendor:
The proposed changes are breaking changes, while there is nothing wrong with breaking changes in hie-bios, I would like to keep the breaking change as minimal as possible.

It seems like you do a lot of parameter passing just to add it to `runGhcCmd`. How about we add it as a paramater to `runGhcCmd`, because otherwise it feels redundant and weird to pass a log function to `runCradle` and pass the log action for `runGhcCmd` via the cradle generation. I think that'd make it more consistent. That, or making `runCradle` to also use the log action passed in to the Cradle creation function.

Otherwise, lgtm, thank you for your contribution!

blackheaven:
No worries, thanks for your feedback, I have pushed a new version integrating them.

fendor (approved):
LGTM, thanks!

PR #378, opened 2022-12-04
Relax lower bounds for aeson and prettyprinter

pepeiborra:
This enables building HLS with stackage LTS 16, which is still important for us. 

I tested this with:

    cabal build --constraint "aeson ==1.4.4.0" --constraint "prettyerinter ==1.6.2" -w ghc-8.8

I will need a Hackage revision please @fendor


fendor (approved):

fendor:
Which hie-bios version you need revisions for?

PR #380, opened 2022-12-13
Allow vector-0.13

Bodigrim:
I checked, it builds fine.


hasufell (approved):

fendor:
I think, vector is even an unused dependency

Issue #25, opened 2019-09-07
Nix support


domenkozar:
I'd like to propose Nix support.

By default, it would use `shell.nix` to enter the environment.

An example default configuration (mostly inspired by stack):

```yaml
nix:
  enable: true
  shell-file: shell.nix
```

matthewess:
Just for reference, in order to fix compatibility between IOHK's haskell.nix infrastructure and ghcide-nix, I'm using a shell script like the following inside of a nix shell to make sure `ghcide` finds things correctly:
```bash
export NIX_GHC="$(which ghc)"
export NIX_GHCPKG="$(which ghc-pkg)"
export NIX_GHC_DOCDIR="$NIX_GHC/../../share/doc/ghc/html"
export NIX_GHC_LIBDIR="$(ghc --print-libdir)"
```

see digital-asset/ghcide#116

Hopefully this can be taken care of by nix integration at the hie-bios level?

acowley:
@matthewess An alternative approach that I've been doing for HIE integration with direnv and emacs is setting those environment variables in a `shellHook` defined in the `shell.nix` for each project so that the paths can be constructed from nix values (I also set `HIE_HOOGLE_DATABASE` in this way). When I gave ghcide a try, I found the same approach worked. I don't think this buys you much if anything, but it's automatic.

freuk:
I similarly re-use the variables from the nix haskell package ghcide is supposed to be used to hack on. Maybe the proposed bit of yaml configuration can allow providing a nix attribute to take the environment variables from.

```
 shellHook = ''
      export NIX_GHC="${haskellPackages.packagename.env.NIX_GHC}"
      export NIX_GHCPKG="${haskellPackages.packagename.env.NIX_GHCPKG}"
      export NIX_GHC_DOCDIR="${haskellPackages.packagename.env.NIX_GHC_DOCDIR}"
      export NIX_GHC_LIBDIR="${haskellPackages.packagename.env.NIX_GHC_LIBDIR}"
'';
```

https://github.com/freuk/very-opinionated-haskell-skeleton/blob/dac5f3ff6b45366e5f4053098dda6d866f7bd0fa/default.nix#L94

turion:
How do the cradle test interplay with nix? I'm currently trying to integrate `hie-bios-0.3` into `nixpkgs` and I get a lot of errors like:

```
    multi-cabal:               FAIL (0.07s)
      Finding Cradle
      Loading Cradle
[...]

PR #382, opened 2022-12-22
Fix import and non-canonical instance warnings

riz0id:
This PR has a few changes that fix GHC warnings when building `lib:hie-bios` and `exe:hie-bios`:

- Removes `Data.Text.Prettyprint.Doc` imports in favor of the `Prettyprinter` module. The `Data.Text.Prettyprint.Doc` module has been deprecated since `prettyprinter-1.7.0`.

- Removes any redundant imports reported by `-Wunused-imports`.

- Encloses the non-canonical `return` definition specified for `CradleLoadResultT` in `MIN_VERSION_base(4,8,0)`, fixing the `-Wnoncanonical-monad-instances` warning.

I expect some of the `prettyprinter` and redundant imports are actually necessary for backward compatibility. I'm happy to add the correct `#if` directives to silence these warning if that is the case.

### DIFF
--- a/src/HIE/Bios/Ghc/Check.hs
+++ b/src/HIE/Bios/Ghc/Check.hs
@@ -19,7 +19,6 @@ import qualified DynFlags as G
 import Control.Exception
 import Control.Monad.IO.Class
 import Colog.Core (LogAction (..), WithSeverity (..), Severity (..), (<&), cmap)
-import Data.Text.Prettyprint.Doc
 
 import HIE.Bios.Ghc.Api
 import HIE.Bios.Ghc.Logger
@@ -27,6 +26,8 @@ import HIE.Bios.Types hiding (Log (..))
 import qualified HIE.Bios.Types as T
 import qualified HIE.Bios.Ghc.Load as Load
 import HIE.Bios.Environment
+
+import Prettyprinter
 
 import System.IO.Unsafe (unsafePerformIO)
 import qualified HIE.Bios.Ghc.Gap as Gap
--- a/src/HIE/Bios/Types.hs
+++ b/src/HIE/Bios/Types.hs
@@ -7,13 +7,13 @@ module HIE.Bios.Types where
 import           System.Exit
 import qualified Colog.Core as L
 import           Control.Exception              ( Exception )
-import           Control.Monad
 import           Control.Monad.IO.Class
 import           Control.Monad.Trans.Class
 #if MIN_VERSION_base(4,9,0)
 import qualified Control.Monad.Fail as Fail
 #endif
-import Data.Text.Prettyprint.Doc
+
+import Prettyprinter
 
 
 data BIOSVerbosity = Silent | Verbose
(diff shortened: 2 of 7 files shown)


src/HIE/Bios/Types.hs
```diff
@@ -122,19 +122,24 @@ instance (Monad m, Applicative m) => Applicative (CradleLoadResultT m) where
                 CradleNone -> pure CradleNone
 
 instance Monad m => Monad (CradleLoadResultT m) where
-  {-# INLINE return #-}
+#if !(MIN_VERSION_base(4,8,0))
   return = CradleLoadResultT . return . CradleSuccess
-  {-# INLINE (>>=) #-}
+  {-# INLINE return #-}
+#endif
```
fendor, line 128: we don't support such old base versions anyway, let's just delete that code.

src/HIE/Bios/Ghc/Check.hs
```diff
@@ -28,6 +27,8 @@ import qualified HIE.Bios.Types as T
 import qualified HIE.Bios.Ghc.Load as Load
 import HIE.Bios.Environment
 
+import Prettyprinter
+
```
fendor, line 31: I think, so far, we used the old style import because @pepeiborra supports, unofficially, HLS for a GHC 8.8.4 stackage snapshot, which only has the old prettyprinter version. Updating this breaks HLS on a GHC 8.8.4. @pepeiborra Is that still relevant?

fendor (changes requested):
Thank you very much, looks good to me, except one nitpick!

fendor (approved):
Thanks, just waiting a few days for pepe, but since they might be on holiday, might take a while

riz0id:
No worries!

riz0id:
@fendor switching to different versions of GHC to test that the version bounds work correctly would probably go faster with nix so I'm going to set that up to save myself some time. I see that an issue #25 is asking for Nix support and a `shell.nix` (I think??). If you would like, I can open a separate PR setting up Nix for `hie-bios` since I'm already going have it ready for finishing this PR.

fendor:
>  If you would like, I can open a separate PR setting up Nix for hie-bios since I'm already going have it ready for finishing this PR.

While I am a nix-user myself, I am a fair bit scared off from maintaining a nix CI. For example, in HLS, nix CI is constantly outdated, which is why we decided to make it a non-blocking CI job... Rendering it relatively useless for nix developers again.
Additionally, I don't quite see where the current set-up is lacking.

So, if you want to "commit" to maintaining the nix CI, I am happy about the contribution, but I don't want to maintain it, unless it is *really* trivial.

riz0id:
Ah, in that case I won't worry about adding Nix in a separate PR (I don't have space to commit to maintaining anything at the moment). 

Out of curiosity though, what do you mean by Nix CI is constantly out of date? I'm not sure I've run into that problem myself. If you could elaborate or possibly link me to an example of an issue that demonstrates the problem I would really appreciate it. I maintain some packages that use Nix CI so knowing about potential pitfalls would be very helpful.

hasufell:
Nix is a time sink. Cache misses that lead to compile failures are not uncommon, despite the claim that it's reproducible.

The ecosystem constantly changes its ways on how to do things, so when you update your recipes, you're likely investing a lot of time again.

riz0id:
I'm not sure how a cache miss could lead to a compile failure unless the Haskell package was just not portable? Do you have a specific example of this happening in mind?

hasufell:
> I'm not sure how a cache miss could lead to a compile failure 

Because if you don't hit the cache, you'll rebuild the entire universe. I've seen build failure in CI numerous times then.

I'm not going to dig those out to make a point though.

riz0id:
Thats fine I just wanted to make sure that your point was not related to Nix.

fendor:
> Out of curiosity though, what do you mean by Nix CI is constantly out of date?

It duplicates the information from .cabal files since nix can't nixify a cabal project automatically, so we have lots of logic for it: https://github.com/haskell/haskell-language-server/blob/master/flake.nix

Also, since it doesn't use the build-plan cabal found, but rather follows a stackage snapshot, you have to customily override versions, jailbreak, not test, etc...

None of this applies to hie-bios directly, it is likely just 20 lines of nix, but imo nix expressions need to be maintained either way, just as nix CI, upgrading nixpkgs pins, etc...

fendor:
If you rebase, this can be merged

jhrcek:
Closing as all of these fixes were done as part of https://github.com/haskell/hie-bios/pull/429

PR #387, opened 2023-01-23
Better support for multi component projects

wz1000:
1. Take the working directory of the current component into account when making DynFlags absolute
2. Allow hie-bios itself to be loaded into a multi component session by making the location of the embedded file
   relative to the working directory of the project.

### DIFF
--- a/src/HIE/Bios/Environment.hs
+++ b/src/HIE/Bios/Environment.hs
@@ -167,9 +167,13 @@ makeDynFlagsAbsolute root df =
         map (Gap.overPkgDbRef makeAbs) (G.packageDBFlags df)
     }
   where
-    makeAbs = case G.workingDirectory df of
-      Nothing -> (root </>)
-      Just fp -> ((root </> fp) </>)
+    makeAbs =
+#if __GLASGOW_HASKELL__ >= 903
+      case G.workingDirectory df of
+        Just fp -> ((root </> fp) </>)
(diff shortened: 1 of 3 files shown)


fendor (approved):
LGTM, thank you!

src/HIE/Bios/Environment.hs
```diff
@@ -159,15 +159,21 @@ addCmdOpts cmdOpts df1 = do
...
   $ df
     { G.importPaths = map makeAbs (G.importPaths df)
     , G.packageDBFlags =
         map (Gap.overPkgDbRef makeAbs) (G.packageDBFlags df)
     }
   where
-    makeAbs = (work_dir </>)
+    makeAbs =
+#if __GLASGOW_HASKELL__ >= 903
+      case G.workingDirectory df of
```
fendor, line 172: So, working directory is relative to the project root directory? And we always know that it is relative and not an absolute path pointing to the correct sub-directory?

wz1000: I think `(</>)` has the right semantics. I can add a comment:

```haskell
λ> "a/b" </> "/absolute"
"/absolute"
λ> "/absolute1" </> "/absolute2"
"/absolute2"
```

fendor: It is fine this way. I just keep forgetting.

PR #403, opened 2023-05-11
Update README for new project-file key

fendor:


README.md
```diff
@@ -318,6 +315,33 @@ cradle:
...
+  cabal:
+    project-file: "./cabal.project.dev"
+```
+
+We can combine the `project-file` field with `components`:
+
+```yaml
+cradle:
+  cabal:
+    project-file: "./cabal.project.dev"
```
fendor, line 333: I just noticed that there is a naming scheme inconsistency, namely `project-file` vs `stackYaml`. We should be consistent, but what would be the better name than `project-file`? `projectFile` looks kind of bad.

fendor: In consistency with `stackYaml`, maybe `cabalProject`?

michaelpj: `cabalProject` sounds reasonable to me.

README.md
```diff
@@ -318,6 +315,33 @@ cradle:
 
 This way we specified which component needs to be compiled given a certain source file for our whole project.
 
+Some projects have multiple `cabal.project` files for multiple versions of ghc or development options. In this case you
+can specify an alternate relative file to use by using the `project-file` option. The path is relative to the
+configuration file.
```
michaelpj, line 320:
```suggestion
`hie.yaml`.
```

michaelpj (approved):

fendor:
For now, I am merging the PR as is, renaming `project-file` to `cabalProject` in a subsequent PR, to keep the separation of Documentation updates and Logic updates.

PR #406, opened 2023-06-11
Add more informative log messages for cradle running

fendor:
Just addig more logging information, including environment variable information.

### DIFF
--- a/src/HIE/Bios/Types.hs
+++ b/src/HIE/Bios/Types.hs
@@ -2,21 +2,56 @@
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE DeriveTraversable #-}
+{-# LANGUAGE OverloadedStrings #-}
 module HIE.Bios.Types where
 
 import           System.Exit
 import qualified Colog.Core as L
 import           Control.Exception              ( Exception )
-import           Control.Monad
 import           Control.Monad.IO.Class
 import           Control.Monad.Trans.Class
 #if MIN_VERSION_base(4,9,0)
 import qualified Control.Monad.Fail as Fail
 #endif
 import Data.Text.Prettyprint.Doc
+import System.Process.Extra (CreateProcess (env, cmdspec), CmdSpec (..))
+import Data.Maybe (fromMaybe)
 
+----------------------------------------------------------------
+-- Environment variables used by hie-bios.
+--
+-- If you need more, add a constant here.
+----------------------------------------------------------------
 
-data BIOSVerbosity = Silent | Verbose
+-- | Environment variable containing the filepath to which
+-- cradle actions write their results to.
+-- If the filepath does not exist, cradle actions must create them.
+hie_bios_output :: String
+hie_bios_output = "HIE_BIOS_OUTPUT"
+
+-- | Environment variable pointing to the GHC location used by
+-- cabal's and stack's GHC wrapper.
+--
+-- If not set, will default to sensible defaults.
+hie_bios_ghc :: String
+hie_bios_ghc = "HIE_BIOS_GHC"
+
+-- | Environment variable with extra arguments passed to the GHC location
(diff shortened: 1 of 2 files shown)


src/HIE/Bios/Types.hs
```diff
@@ -48,14 +83,32 @@ data ActionName a
...
   pretty (LogAny s) = pretty s
   pretty (LogProcessOutput s) = pretty s
+  pretty (LogProcessRun fp args) = pretty fp <+> pretty (unwords args)
+  pretty (LogCreateProcessRun cp) =
+    vcat $ 
+      [ case cmdspec cp of
+          ShellCommand sh -> pretty sh
+          RawCommand cmd args -> pretty cmd <+> pretty (unwords args)
+      ]
+      <>
```
michaelpj, line 103:
```suggestion
      line
```

src/HIE/Bios/Types.hs
```diff
@@ -48,14 +83,32 @@ data ActionName a
...
+      if null envText
+        then []
+        else
+          [ indent 2 $ "Environment Variable"
```
michaelpj, line 107:
```suggestion
          [ indent 2 $ "Environment Variables"
```

src/HIE/Bios/Types.hs
```diff
@@ -48,14 +83,32 @@ data ActionName a
...
   pretty (LogProcessOutput s) = pretty s
+  pretty (LogProcessRun fp args) = pretty fp <+> pretty (unwords args)
+  pretty (LogCreateProcessRun cp) =
+    vcat $ 
+      [ case cmdspec cp of
+          ShellCommand sh -> pretty sh
+          RawCommand cmd args -> pretty cmd <+> pretty (unwords args)
+      ]
+      <>
+      if null envText
```
michaelpj, line 104: could put the `indent 2` outside the `if` if you want

src/HIE/Bios/Types.hs
```diff
@@ -48,14 +83,32 @@ data ActionName a
...
+        then []
+        else
+          [ indent 2 $ "Environment Variable"
+          , indent 2 $ vcat envText
+          ]
+    where
+      envText = map (indent 2 . pretty) $ prettyProcessEnv cp
```
michaelpj, line 111:
```suggestion
      envText = indent 2 $ vcat $ map pretty $ prettyProcessEnv cp
```
?

src/HIE/Bios/Types.hs
```diff
@@ -191,3 +244,24 @@ data ComponentOptions = ComponentOptions {
   -- changes the options that a Cradle may return, thus, needs reload
   -- as soon as these files are created.
   } deriving (Eq, Ord, Show)
+
+
+-- ------------------------------------------------
+
+-- | Prettify 'CmdSpec', so we can show the command to a user
+prettyCmdSpec :: CmdSpec -> String
```
michaelpj, line 252: These aren't returning `Doc` because they're being used by other stuff, right?

michaelpj (approved):
LGTM

src/HIE/Bios/Types.hs
fendor: :scream: thanks!

src/HIE/Bios/Types.hs
fendor: doesnt work, because then `null envText` fails

src/HIE/Bios/Types.hs
fendor: That adds an extra new line in the case where we have no environment variables.

src/HIE/Bios/Types.hs
fendor: Yes, exactly.

src/HIE/Bios/Types.hs
michaelpj: You can do `vcat [doc1, doc2]` if you like. Just that this one will try to lay it out on one line, which you probably don't want

PR #407, opened 2023-06-16
Rename project-file to cabalProject in hie.yaml

fendor:
This is consistent with the name 'stackYaml' for stack cradles. Unfortunate naming, but consistency beats the potential generalisation.

Follow up to #403


fendor:
This should be done before the incoming release of hie-bios.

PR #415, opened 2023-10-09
Accept directories in 'findCradle'

fendor:
See https://github.com/haskell/haskell-language-server/pull/3830 and https://github.com/haskell/haskell-language-server/issues/3074

### DIFF
--- a/src/HIE/Bios/Cradle.hs
+++ b/src/HIE/Bios/Cradle.hs
@@ -77,10 +77,17 @@ import Text.ParserCombinators.ReadP (readP_to_S)
 
 ----------------------------------------------------------------
 
--- | Given root\/foo\/bar.hs, return root\/hie.yaml, or wherever the yaml file was found.
+-- | Given @root\/foo\/bar.hs@, return @root\/hie.yaml@, or wherever the yaml file was found.
+--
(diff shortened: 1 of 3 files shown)


src/HIE/Bios/Cradle.hs
```diff
@@ -77,10 +77,17 @@ import Text.ParserCombinators.ReadP (readP_to_S)
...
+-- | Given @root\/foo\/bar.hs@, return @root\/hie.yaml@, or wherever the yaml file was found.
+--
+-- Note, 'findCradle' used to **not** work for directories and required a Haskell file.
+-- This has been fixed since @0.14.0@.
+-- However, 'loadCradle' and 'loadImplicitCradle' still require a Haskell
+-- source file and won't work properly with a directory parameter.
 findCradle :: FilePath -> IO (Maybe FilePath)
 findCradle wfile = do
-    let wdir = takeDirectory wfile
+    wdir <- doesDirectoryExist wfile >>= \case
```
michaelpj, line 88: I guess this will behave a bit weirdly for directories that don't exist. I guess we have no way of knowing if a path points to a directory or not without checking on the filesystem...

michaelpj (approved):
Test?

PR #422, opened 2023-11-14
improve implicit cradle

wz1000:
Improve implicit cradles.

Try to match the behaviour of implicit-hie-cradle as closely as possible.

Elminates the need for implicit-hie-cradle and consolidates cradle selection
logic into hie-bios.

1. Prefer cabal over stack (all else being equal)
2. Take work directories into account when making a decision
3. Use `implicit-hie` to generate stack multi cradles

In particular, this differs from `implicit-hie-cradle` in that it uses default
(`cradle: cabal:`) cradles for cabal.

However, we still generate implicit multi-cradles for stack projects by parsing
the stack.yaml and *.cabal files

### DIFF
--- a/hie-bios.cabal
+++ b/hie-bios.cabal
@@ -172,6 +172,7 @@ Library
                         exceptions           ^>= 0.10,
                         cryptohash-sha1      >= 0.11.100 && < 0.12,
                         directory            >= 1.3.0 && < 1.4,
+                        implicit-hie         >= 0.1.4.0 && < 0.1.5,
                         filepath             >= 1.4.1 && < 1.5,
                         time                 >= 1.8.0 && < 1.13,
                         extra                >= 1.6.14 && < 1.8,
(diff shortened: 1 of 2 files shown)


wz1000:
I've moved this into HLS proper: https://github.com/haskell/haskell-language-server/pull/3462/commits/178e54f8b2367ca47619081677d127fd3fa94c61

michaelpj:
Let's kill it!

jhrcek:
@fendor  If I understand this correctly, the logic from this PR now lives in hls: https://github.com/haskell/haskell-language-server/blob/master/ghcide/session-loader/Development/IDE/Session/Implicit.hs

So can this PR be closed then?

fendor:
I think we wanted to remove the implicit cradle logic since we don't need it any more?

jhrcek:
I can work on that if you give me some pointers, but in any case this PR looks like it's no longer relevant.

fendor:
To remove the implicit cradle stuff, just start from https://github.com/haskell/hie-bios/pull/422/files#diff-58ae540d147057c94682a01b61de564caf1eb641fac8d2a3f29ea1f29d673ac4L316 and delete stuff until everything's green again :D

fendor:
I think the implicit cradle stuff has been merged into HLS directly, so closing this.

PR #433, opened 2024-04-16
Add Loading Style option to 'runAction'

fendor:


src/HIE/Bios/Cradle.hs
```diff
@@ -284,15 +284,22 @@ resolvedCradlesToCradle logger buildCustomCradle root cs = mdo
...
     ConcreteStack t -> stackCradle l root (stackComponent t) (projectConfigFromMaybe root (stackYaml t))
     ConcreteBios bios deps mbGhc -> biosCradle l root bios deps mbGhc
     ConcreteDirect xs -> directCradle l root xs
     ConcreteNone -> noneCradle
     ConcreteOther a -> buildCustomCradle a
+  where
+    -- Add a log message to each loading operation.
+    addLoadStyleLogToCradleAction crdlAct = crdlAct
+      { runCradle = \fp ls -> do
+          l <& LogCradleLoadStyle (T.pack $ show $ actionName crdlAct) ls `WithSeverity` Debug
```
michaelpj, line 300: I think this going to end up being a bit odd, right? We're logging the user's _requested_ loading style, but then we might not be able to honour it. e.g. if the user requests multi-component but they have an old cabal. Or if they pick auto who knows what they'll get. So we need to log something about the load style that we actually _use_ also.

src/HIE/Bios/Cradle.hs
```diff
@@ -787,34 +794,39 @@ cabalAction
...
+    (Just cabal, Just ghc)
+      -- Multi-component supported from cabal-install 3.11
+      -- and ghc 9.4
+      | ghc   >= makeVersion [9,4]
+      , cabal >= makeVersion [3,11]
+      -> case loadStyle of
+        SingleComponent -> pure [fromMaybe (fixTargetPath fp) mc]
+        -- Start a multi-component session with all the old files
+        LoadWithContext fps -> do
+          liftIO $ l <& LogCabalMultiComponentSupport `WithSeverity` Info
```
michaelpj, line 814: Okay, I see we're doing something here.

I guess maybe what I'd be inclined to do would be something like:
```
log $ RequestedLoadStyle userLoadStyle
determinedLoadStyle <- case (Just cabal, Just ghc) of ...
log $ DeterminedLoadStyle determinedLoadStyle
case determinedLoadStyle of ...
```

michaelpj: Also I don't think you're going to get anything useful if you pick the multi-component load style with a stack cradle.

michaelpj (commented):
Possibly stupid question, but would it make more sense to just consider this a different cradle type? It's not clear that it really is an orthogonal setting, e.g. it makes no sense if you set it and you have a stack or direct cradle. Or maybe you should get an error in that case? But a separate cradle type might work nicely. Then in HLS/hie-bios we can give people a choice of direct/stack/cabal/cabal-multi.

michaelpj:
Also, if we did my suggestion about making this a cradle type, then exposing that in HLS would naturally fix https://github.com/haskell/haskell-language-server/issues/3832

src/HIE/Bios/Cradle.hs
fendor: Yeah, but we want all cradles to have the same interface. Otherwise, it gets tricky real quickly and this model has served us well, so far.

michaelpj: Right, I just want to get some kind of clear log output indicating that the option is doing nothing in that case.

fendor: Indeed, that's a good idea! Changed the logging to log in the case of `stack` and `bios` cradle that the load style is not supported.

src/HIE/Bios/Cradle.hs
```diff
@@ -779,42 +788,55 @@ cabalGhcDirs l cabalProject workDir = do
...
+      -- and ghc 9.4
+      | LoadWithContext _ <- loadStyle ->
+          if ghc >= makeVersion [9,4] && cabal >= makeVersion [3,11]
+            then pure loadStyle
+            else do
+              liftIO $ l <& WithSeverity
+                (LogLoadWithContextUnsupported "cabal"
+                  $ Just "cabal or ghc version is too old. We require `cabal >= 3.11` and `ghc >= 9.4`."
+                )
+                Info
```
michaelpj, line 818: Warning, I would think? "You asked to do something but we can't" seems like classic warning territory

src/HIE/Bios/Types.hs
```diff
@@ -116,11 +120,31 @@ instance Pretty Log where
           ]
     where
       envText = map (indent 2 . pretty) $ prettyProcessEnv cp
+  pretty (LogRequestedCradleLoadStyle crdlName ls) =
+    "Request loading" <+> pretty crdlName <+> "cradle using" <+> case ls of
+      SingleComponent -> "Single Component Strategy"
+      LoadWithContext fps -> "Multiple Components Strategy:" <> line <> indent 4 (pretty fps)
```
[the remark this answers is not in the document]
michaelpj: It seems like from the `hie-bios` perspective the truth is that it's either one _file_ or _many files_, right? `hie-bios` doesn't particularly judge whether or not they're in different components or anything. Just that you can ask for flags to load one file, or a bunch of files.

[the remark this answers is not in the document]
fendor: Yeah, that is true. But logs are still kind of user facing, so I thought, it might be better to use the terminology we would use in HLS.

Should we uniformly use the `hie-bios` perspective here?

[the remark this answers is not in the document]
michaelpj: I think it makes `hie-bios` more confusing if we're not consistent. But maybe we're just a bit confused about what `hie-bios` is saying it will do? Or perhaps we can make the message informative enough to cover both: "Requested to load cradle using all open files (multi-component)" or something?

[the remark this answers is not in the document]
fendor: that's a good suggestion, updated!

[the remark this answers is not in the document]
fendor: I also renamed the `LoadStyle` constructors to reflect this more accurately.

Issue #405, opened 2023-06-10
Progress reporting


michaelpj:
At the moment we don't give good feedback to the user when the build tool is building stuff. _Ideally_, we would show something like "Building dependencies 54/500" or something via progress reporting.

However this is pretty hard to do well since we just shell out to the build tool and don't have a proper interface. Nonetheless, a hacky implementation in terms of parsing the stdout or something would be really nice for users and so I think it's probably worth it.

We'd need two things:
1. An interface for progress reporting for cradles
2. An implementation of some kind of progress reporting for, say, cabal

PR #436, opened 2024-06-10
Add progress reporting support to cabal building

palas:
This PR adds a tiny parser to `readCabalProcessWithProgress` and it takes a new reporting function as a parameter that can be used to notify the caller about the progress of the building (how many dependencies to build and how many have been built so far). The motivation is to support HLS on this issue: https://github.com/haskell/hie-bios/issues/405


src/HIE/Bios/Cradle.hs
```diff
@@ -1212,6 +1227,43 @@ readProcessWithOutputs outputNames l workDir cp = flip runContT return $ do
...
+      reportProgress (Just reporter) str cps@(CabalParserBuilding { numPackagesCompleted = numPackages
+                                                                  , numPackagesToBuild = totalPackages
+                                                                  }) =
+        case str of
+           'C':'o':'m':'p':'l':'e':'t':'e':'d':' ':' ':' ':' ':_ -> do
+             reporter (CompilationProgress { numPackagesToCompile = totalPackages
+                                           , numPackagesCompiled = numPackages + 1
+                                           })
+             pure (cps { numPackagesCompleted = numPackages + 1 }, str)
+           _ -> pure (cps, str)
```
fendor, line 1250: I think running a process shouldn't be coupled this tightly with a particular build tool. How about we pull this out and pass the progress reporter to the function?

src/HIE/Bios/Cradle.hs
```diff
@@ -566,8 +573,8 @@ cabalCradle l cs wdir mc projectFile
 -- to the custom ghc wrapper via 'hie_bios_ghc' environment variable which
 -- the custom ghc wrapper may use as a fallback if it can not respond to certain
 -- queries, such as ghc version or location of the libdir.
-cabalProcess :: LogAction IO (WithSeverity Log) -> CradleProjectConfig -> FilePath -> String -> [String] -> CradleLoadResultT IO CreateProcess
-cabalProcess l cabalProject workDir command args = do
+cabalProcess :: CompilationProgressReporter -> LogAction IO (WithSeverity Log) -> CradleProjectConfig -> FilePath -> String -> [String] -> CradleLoadResultT IO CreateProcess
+cabalProcess _ l cabalProject workDir command args = do
```
fendor, line 577: Why unused variable?

fendor (changes requested):
Hi, thank you for your PR!

I like the improved progress reporting, a couple of suggestions!

src/HIE/Bios/Cradle.hs
```diff
@@ -1158,19 +1166,18 @@ getCleanEnvironment = do
...
--- * The logging function is called every time the process emits anything to stdout or stderr.
--- it can be used to report progress of the process to a user.
--- * The process is executed in the given directory.
-readProcessWithOutputs
-  :: Outputs  -- ^ Names of the outputs produced by this process
+data CabalParserState = CabalParserToBuild { numPackagesDeclared :: Int }
+                      | CabalParserBuilding { numPackagesCompleted :: Int, numPackagesToBuild :: Int }
+
+-- | Same as 'readProcessWithOutputs' but reports process when running cabal build
+readCabalProcessWithProgress
```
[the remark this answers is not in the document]
palas: Good point, not sure how I missed that. Fixed here: https://github.com/haskell/hie-bios/pull/436/commits/92a2359ab0c13e9f6106d2e4db5c70e29b0591cd

src/HIE/Bios/Cradle.hs
palas: Good point, not sure how I missed that. Fixed here: https://github.com/haskell/hie-bios/commit/92a2359ab0c13e9f6106d2e4db5c70e29b0591cd

src/HIE/Bios/Cradle.hs
palas: Good point. It is much more elegant if I just pass an accumulator function, it also highlighted that I was passing the string through the function unnecessarily. Fixed here: https://github.com/haskell/hie-bios/pull/436/commits/6d71b932eab6b38c6f07d6b4a648cdea7e866fe2

src/HIE/Bios/Cradle.hs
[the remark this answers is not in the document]
palas: Ah, this makes more sense after addressing the other comment: https://github.com/haskell/hie-bios/pull/436/commits/6d71b932eab6b38c6f07d6b4a648cdea7e866fe2

soulomoon:
Look like a good feature to me. It is possible to add a unittest for it ?

guibou:
There had been a rant on reddit which complains that HLS gives poor feedback when building dependencies (don't reed the reddit topic, they use strong words which does not really motivate me to fix their problem, but the problem is actually genuine).

Maybe that's a low hanging fruit which will help many users (I've been confused too a few times).

https://www.reddit.com/r/haskell/comments/1v8zs79/rant_haskell_tooling_specifically_hls_is_just/

(I'm writing a comment here so maybe I'll have a look)

PR #437, opened 2024-07-02
Add support for bios multi-cradles

iteratee:
We currently don't have a way to ask a bios cradle if it supports multiple components. So we just trust that if the GHC version is new enough and multiple component loading is requested, that the bios program supports it.


fendor:
Hi, thank you for your PR! 

I am worried that passing multiple filepaths to the bios command might break existing bios scripts, since they may assume it contains only exactly one filepath. We could pass in the extra filepaths in a separate environment variable `HIE_BIOS_EXTRA_ARGS` , then bios scripts cannot break accidentally. What do you think?

iteratee:
> Hi, thank you for your PR!
> 
> I am worried that passing multiple filepaths to the bios command might break existing bios scripts, since they may assume it contains only exactly one filepath. We could pass in the extra filepaths in a separate environment variable `HIE_BIOS_EXTRA_ARGS` , then bios scripts cannot break accidentally. What do you think?

I'm very sorry that I forgot about this.

I've been using my own patched version of hie-bios since I posted this last year.

I went with `HIE_BIOS_MULTI_ARG`

The files are separated by `'\x1f'`, the record separator character. Null won't work because environment variables are null terminated. It's possible to put a record separator in a file name, but I can accept that limitation.

Let me know what you think.

Kyle.

fendor:
> The files are separated by '\x1f', the record separator character. Null won't work because environment variables are null terminated. It's possible to put a record separator in a file name, but I can accept that limitation.

What we usually do is to generate a temporary file, pass this file as the value of the environment variable, and write the filepaths to this temporary file instead of the environment variable directly.
It should be reasonably simple to do that, I think and should even lift the last restriction, right?

iteratee:
> > The files are separated by '\x1f', the record separator character. Null won't work because environment variables are null terminated. It's possible to put a record separator in a file name, but I can accept that limitation.
> 
> What we usually do is to generate a temporary file, pass this file as the value of the environment variable, and write the filepaths to this temporary file instead of the environment variable directly. It should be reasonably simple to do that, I think and should even lift the last restriction, right?

Yes, that should be reasonably simple, and then we can null separate the files. I'll see what I can do.

iteratee:
I got that working, and pushed up the result if you want to take a look.

Issue #440, opened 2024-09-25
Batch loader to better support laoding of multiple home units at once


soulomoon:
Originates from https://github.com/haskell/haskell-language-server/issues/4381
Batch loading api is needed for performance.
I assume something like this:
```
loadCradles ::  [FilePath] -> IO [(Cradle Void, [FilePath])]
-- CradleAction
runCradleMultiFiles :: [FilePath] -> LoadStyle -> IO (CradleLoadResult ComponentOptionsMultiResult)
```

soulomoon:
The current status for a single file:

1. find the cradle location for a file.
2. load `hie.yaml` or implicit one as `cradle config`.
3. resolve `cradle config` for the file to get the`Cradle`. (Only here we know what build tools are used for the file)
4. `runCradle` to run the `cradleAction` in `Cradle` and obtain the arguments`ghc` needed to compile the file

Since only `cabal` cradle can do batch load. (Multiple cabal files?)
To minimize changes need to be made, internal changes need to be made to `cabalActions`.
Because we need to merge multiple `cabalAction`. Current implementation is not possible.

Then the batch load workflow can be:

* For each file
    * find the cradle location
    * load the `hie.yaml` to `cradle config`
    * resolve get `Cradle` for the file.
* Merge the `Cabal` cradles, leave others.
* `runCradle` for each batch of files

soulomoon:
loadStyle already fullfill batch load all the modules for a single cradle project. No need to further load.

PR #445, opened 2024-10-24
Batch load

soulomoon:
Attempt for #440


soulomoon:
close infavor of https://github.com/haskell/hie-bios/pull/453

PR #458, opened 2025-04-18
Use cabal path when available

alexfmpe:
Before 
```
$ time cabal run hie-bios -- debug hie-bios.cabal
real	0m5.409s
user	0m3.587s
sys	0m1.738s
```
After
```
$ time cabal run hie-bios -- debug hie-bios.cabal
real	0m4.150s
user	0m2.652s
sys	0m1.377s
```


alexfmpe:
The presence of 
https://github.com/haskell/hie-bios/blob/6ccbb45ae5ec4e3bde8e7eafc0a67bba33d888bd/src/HIE/Bios/Cradle.hs#L606
doesn't seem to make a difference with the `cabal path` flow. 
Is it worth making it conditional?

alexfmpe:
@fendor Even after the original changes to `cabalProcess`, I noticed in the logs a few 
```
[Debug] cabal --builddir=<blabla> v2-exec --with-compiler <blabla> --with-hc-pkg <blabla> ghc -v0 -- --numeric-version
[Debug] cabal --builddir=<blabla> v2-exec --with-compiler <blabla> --with-hc-pkg <blabla> ghc -v0 -- --print-libdir
```
and was able to shave off more time by doing the same cabal path trick there.

Just like you pointed out in https://github.com/haskell/haskell-language-server/issues/3578#issuecomment-1747452570, I see a lot of repeated commands in the logs (and with even more repetition now that I'm doing extra calls to `cabal --version`) and think things can be made faster still by fixing that but it looks like that'll require less local changes so leaving it for a future PR

alexfmpe:
Am I reading the above logs wrong or were we doing
`cabal exec -v0 -- ghc --print-libdir` 
to build the wrapper to do 
`cabal --builddir=<blabla> v2-exec --with-compiler <blabla> --with-hc-pkg <blabla> ghc -v0 -- --print-libdir`

fendor:
> Am I reading the logs wrong 

You are reading it correctly! We are essentially building a custom GHC shim, that requires us to find the `ghc-pkg` location, the global package db and the libdir.

> doesn't seem to make a difference with the cabal path flow.

I conjecture `--hc-pkg` flag might not be needed any more, so I would like to get rid of it if possible.
Aren't the code paths different enough already, s.t. it barely hurts to have another conditional :)

> I wonder if something should be a fatal error or at least a warning rather than silently fallback to ...

Yeah, logging a warning sounds good!

---

These performance improvements sound much bigger than I had anticipated, awesome work! I away for the weekend, but will review ASAP.

fendor:
I think I changed to settings and you should be able to run the CI workflow now!

alexfmpe:
> We are essentially building a custom GHC shim, that requires us to find the ghc-pkg location, the global package db and the libdir.

Sure, but are those two `--print-libdir` supposed to be different ?

> I think I changed the settings, and you should be able to run the CI workflow now!

Duplicated CI matrix by also running everything with cabal 3.14, but maybe this is overkill. An alternative would be testing 9.12 with both cabals on all three systems

Though...all the cabal 3.14 jobs fail, as do cabal 3.10 on windows for 9.4 and 9.6?

fendor:
> Sure, but are those two --print-libdir supposed to be different ?

Not sure to what you are referring to exactly, but I think they shouldn't be different, it should be essentially `$(cabal path --output-format=json | jq -r '.compiler.path') --print-libdir`

mpickering:
Could someone perhaps update the MR description and commit message to describe the proposed change in this patch?

fendor:
@alexfmpe The test that fails is genuine, you need to pass the `CradleProjectConfig` to `cabalPathCompilerPath` as well and use it in the `cabal path --project-file` invocation.

src/HIE/Bios/Cradle.hs
```diff
@@ -796,6 +811,17 @@ cabalGhcDirs l cabalProject workDir = do
   where
     projectFileArgs = projectFileProcessArgs cabalProject
 
+cabalPathCompilerPath :: LogAction IO (WithSeverity Log) -> FilePath -> CradleLoadResultT IO (Maybe FilePath)
+cabalPathCompilerPath l workDir = do
+  liftIO (getCabalVersion l workDir) >>= \case
+    Just cabal_version | cabal_version >= makeVersion [3,14] -> do
+      compiler_info <- readProcessWithCwd_ l workDir "cabal" ["path", "--compiler-info"] ""
+      let compiler_path = asum $ stripPrefix "compiler-path: " <$> lines compiler_info
```
hasufell, line 819: Why not use `--output-format=json` and then access the keys properly? I'm not sure `key-value` is very suitable for machine reading?

fendor: Thanks for pointing this out, I also think we should the json output format.

alexfmpe: Sure. Do I use `aeson` or something else?
I'm only familiar with that one but would prefer a *fast* parser. We're already slow as is

alexfmpe:
> @alexfmpe The test that fails is genuine, you need to pass the `CradleProjectConfig` to `cabalPathCompilerPath` as well and use it in the `cabal path --project-file` invocation.

Hmm I don't have that failure locally, what am I doing wrong? That is, how do I reproduce?

fendor:
@alexfmpe You need to install `ghc-8.10.7` locally. It is skipped by default if `ghc-8.10.7` is not installed.
I suppose that's a bad default?

alexfmpe:
Aaaah. Yeah maybe it should be on by default and if anything in that testgroup fails, then it could tell you a flag to disable those tests? Not familiar with how often false positives/negatives occur here so no strong opinion

alexfmpe:
> Not sure to what you are referring to exactly,

Just wondering about the reason for grabbing libdir to build a wrapper for grabbing libdir

src/HIE/Bios/Cradle.hs
```diff
@@ -796,6 +815,23 @@ cabalGhcDirs l cabalProject workDir = do
   where
     projectFileArgs = projectFileProcessArgs cabalProject
 
+cabalPathCompilerPath :: LogAction IO (WithSeverity Log) -> FilePath -> CradleProjectConfig -> CradleLoadResultT IO (Maybe FilePath)
+cabalPathCompilerPath l workDir projectFile = do
+  liftIO (getCabalVersion l workDir) >>= \case
+    Just cabal_version | cabal_version >= makeVersion [3,14] -> do
```
fendor, line 821: I think it makes sense to extract this check into its own function with a name and documentation, similar to `isCabalMultipleCompSupported`.

src/HIE/Bios/Cradle.hs
```diff
@@ -796,6 +815,23 @@ cabalGhcDirs l cabalProject workDir = do
...
+      let
+        args = ["path", "--output-format=json"] <> projectFileProcessArgs projectFile
+        bs = BS.fromStrict . T.encodeUtf8 . T.pack
+        parse_compiler_path = Aeson.parseEither ((.: "compiler") >=>  (.: "path")) <=< Aeson.eitherDecode
+
+      compiler_info <- readProcessWithCwd_ l workDir "cabal" args ""
+      case parse_compiler_path (bs compiler_info) of
+        Left err -> do
+          liftIO $ l <& WithSeverity (LogAny $ "Could not parse json output of 'cabal path': " <> T.pack err) Warning
```
fendor, line 830: A proper log constructor would be good!

src/HIE/Bios/Cradle.hs
```diff
@@ -796,6 +815,23 @@ cabalGhcDirs l cabalProject workDir = do
   where
     projectFileArgs = projectFileProcessArgs cabalProject
 
+cabalPathCompilerPath :: LogAction IO (WithSeverity Log) -> FilePath -> CradleProjectConfig -> CradleLoadResultT IO (Maybe FilePath)
```
fendor, line 818: A little bit more documentation would be appreciated!

fendor (approved):
LGTM, a couple of nitpicks!

fendor:
Pretty solid performance improvements on the test suite! Looks like around 20s across all platforms. 

Great work!

src/HIE/Bios/Cradle.hs
alexfmpe: Wanted to make it consistent with `isCabalMultipleCompSupported` and just realised `ProgramVersions` was actually available on both codepaths that reach here. This cut redundant invocations to `cabal --version`, at the tune of 50ms each on my machine.

src/HIE/Bios/Cradle.hs
alexfmpe: Renamed so it's less ambiguous.
Really not sure what else to say about it besides "calls cabal path to get compiler's path".

fendor:
> For whatever reason the cabal --exec route is way slower on my 7 years old x86_64 linux than on my M1

`cabal exec` does a lot of work, I think it runs the solver, potentially downloads tarballs, checks whether everything is up-to-date, etc... It makes sense that this is even more expensive on older machines.

The test suite gains are quite significant, considering only a subset of tests is using `cabal`, so this will be quite noticeable in day-to-day work, especially with VS Code.

tomsmeding:
> The presence of
> 
> https://github.com/haskell/hie-bios/blob/6ccbb45ae5ec4e3bde8e7eafc0a67bba33d888bd/src/HIE/Bios/Cradle.hs#L606
> 
> doesn't seem to make a difference with the `cabal path` flow.
> Is it worth making it conditional?

It does make a difference if the compiler version set in the project (using `with-compiler:` in `cabal.project` or `cabal.project.local`) is not the ambient `ghc` version.

```sh
$ cat fakeghc.sh
#!/usr/bin/env bash
ghc-9.10.2 "$@"

$ cat fakeghcpkg.sh
#!/usr/bin/env bash
ghc-pkg-9.10.2 "$@"

$ ghc --numeric-version  # note: different!
9.8.4

$ cabal repl --with-compiler ./fakeghc.sh
Error: [Cabal-4000]
Version mismatch between ghc and ghc-pkg: ./fakeghc.sh is version 9.10.2 /home/tom/.ghcup/bin/ghc-pkg is version 9.8.4


$ cabal repl --with-compiler ./fakeghc.sh --with-hc-pkg ./fakeghcpkg.sh
Build profile: -w ghc-9.10.2 -O1
In order, the following will be built (use -v for more details):
 - temphs-0.1.0.0 (interactive) (exe:temphs) (first run)
Preprocessing executable 'temphs' for temphs-0.1.0.0...
GHCi, version 9.10.2: https://www.haskell.org/ghc/  :? for help
Loaded GHCi configuration from /home/tom/git/dotfiles/.ghci
[1 of 2] Compiling Main             ( app/Main.hs, interpreted )
Ok, one module loaded.
*Main>
Leaving GHCi.

$
```

The upshot is that if you have:
- cabal 3.14
- a new HLS (2.11.0.0, I guess)
- `ghc` set to version A
- a `cabal.project` with `with-compiler: ghc-B` where A ≠ B
then HLS is broken because all you get in every file is:
```
Failed to run ["cabal","v2-repl","/home/tom/tmp/temphs/app/Main.hs"] in directory "/home/tom/tmp/temphs". Consult the logs for full command and error.
Failed command: cabal --builddir=/home/tom/.cache/hie-bios/dist-temphs-8641907d45cb211ac5ee400b50e468b9 v2-repl --with-compiler /home/tom/.cache/hie-bios/wrapper-b54f81dea4c0e6d1626911c526bc4e36 /home/tom/tmp/temphs/app/Main.hs
Build profile: -w ghc-9.8.4 -O1
In order, the following will be built (use -v for more details):
 - temphs-0.1.0.0 (interactive) (exe:temphs) (first run)
Configuring executable 'temphs' for temphs-0.1.0.0...

Configuration is affected by the following files:
- cabal.project
Error: [Cabal-4000]
Version mismatch between ghc and ghc-pkg: /home/tom/.cache/hie-bios/wrapper-b54f81dea4c0e6d1626911c526bc4e36 is version 9.10.2 /home/tom/.ghcup/bin/ghc-pkg is version 9.8.4

Error: [Cabal-7125]
Failed to build exe:temphs from temphs-0.1.0.0. The failure occurred during the configure step.



Process Environment:
HIE_BIOS_GHC: /home/tom/.ghcup/bin/ghc-9.10.2
HIE_BIOS_GHC_ARGS: -B/home/tom/.ghcup/ghc/9.10.2/lib/ghc-9.10.2/lib
```

Fortunately, there is a simple workaround (downgrade to cabal 3.12), but I think this counts as a bug in HLS. Should I create a separate issue for this?

alexfmpe:
@tomsmeding oops, see https://github.com/haskell/hie-bios/pull/465

PR #462, opened 2025-04-25
Fix CI caches

fendor:
The cache key needs to be extended to account for the `.cabal` version, otherwise we poison the caches of each CI job.


alexfmpe (approved):
Oops

PR #464, opened 2025-05-26
fix: rewrite findFileUpwards as a non predicate function

guibou:
- `findFileUpwards` is renamed to `findFileUpwardsPredicate`.
- `findFileUpwards x dir` now behaves as `findFileUpwardsPredicate (x ==) dir` but can be dramatically faster.
- Rewrote all uses cases to use the `findFileUpwards` when the filename is already known.

The problem with the version with a predicate is that it needs to load all the files contained in each directories it traverses in order to apply the predicate. Even if these directories contains really few files, that's still slower than a direct test with `doesFileExists`.

But the situation can became really awful if the traversed directories contains a huge amount of file.

And huge can be any version of huge. For example, when working on the startup time of haskell-language-server on my work repository, see https://github.com/haskell/haskell-language-server/issues/4598 for the discussion, I observed that hls was spending 20s in `findFileUpwards`: hls was trying to locate the `hie.yaml` file starting from a repository which is stored in `/nix/store`, which contains hundreds of thousands of entries on my computer (and could actually be more).

This fix removes `20s` of the startup time of HLS on my work project and may do the same for anyone who have the chance of using HLS on a file which is stored in their `/nix/store`.

Note that unfortunately the cabal cradle still suffers from this issues.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,5 +1,8 @@
 # ChangeLog hie-bios
 
+* Rename `findFileUpwards` as `findFileUpwardsPredicate` and implement
+  `findFileUpward` which does have better performance guarantees.
+
 ## 2025-04-25 - 0.15.0
 
 * Use consistent logging for test cases [#461](https://github.com/haskell/hie-bios/pull/461)
--- a/src/HIE/Bios/Cradle.hs
+++ b/src/HIE/Bios/Cradle.hs
@@ -364,7 +364,7 @@ yamlConfig fp = do
   return (configDir </> configFileName)
 
 yamlConfigDirectory :: FilePath -> MaybeT IO FilePath
-yamlConfigDirectory = findFileUpwards (configFileName ==)
+yamlConfigDirectory = findFileUpwards configFileName
 
 readCradleConfig :: Yaml.FromJSON b => FilePath -> IO (CradleConfig b)
 readCradleConfig yamlHie = do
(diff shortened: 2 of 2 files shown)


fendor (approved):
LGTM, good find!

fendor:
> Oh, I see, it is because cabal with none-deterministic name *.cabal

Perhaps we can improve it on average with an educated guess? Many `*.cabal` files are named `<basename `pwd`>.cabal` and we could try that name first.

guibou:
Nother possible ugrade, which won't change the `O` behavior, but will change the constant by a large factor would be to do the directory walk by ourselfs.

Right now, we have:

```haskell
findFile :: (FilePath -> Bool) -> FilePath -> IO [FilePath]
findFile p dir = do
  b <- doesDirectoryExist dir
  if b then getFiles >>= filterM doesPredFileExist else return []
  where
    getFiles = filter p <$> getDirectoryContents dir
    doesPredFileExist file = doesFileExist $ dir </> file
```

This does list the complete directory content in memory (as `FilePath`), so it is potentielally millions (if more) of `(:)` and `Char` in RAM.

Instead, the directory can be walked in an `O(1)` memory / alloc pattern.

Bodigrim:
> Nother possible ugrade, which won't change the O behavior, but will change the constant by a large factor would be to do the directory walk by ourselfs.

You can use `directory-ospath-streaming` to improve constant factor and reduce memory residency to O(1). Its recursive walk uses `unsafeInterleaveIO` avoiding allocation of entire list in memory.

PR #465, opened 2025-06-03
Always specify '--with-hc-pkg' with cabal

alexfmpe:
Should fix https://github.com/haskell/hie-bios/pull/458#issuecomment-2931984794
I'd expect CI to catch this but something's off with the test suite.
I can reproduce manually by adding a `with-compiler: ghc-B` to hie-bios's own cabal.project and running `hie-bios debug` against the .cabal file.
However, I can't seem to do the same for `cabal-with-ghc` and `cabal-with-ghc-and-project` tests even if I try something like
```diff
- initCradle "src/MyLib.hs"
+ let fp = "src/MyLib.hs"
+ initCradle fp
+ liftIO . getCompilerOptions fp LoadFile =<< askCradle
```


fendor (approved):
LGTM, it would be great to make sure we have a regression test so this doesn't occur again...

fendor:
No, I will have to take a closer look

PR #466, opened 2025-06-17
Load session using cabal's `--with-repl` command 

### LINKED ISSUE #467: Cabal cradle needs to use `--with-repl` option
In https://github.com/haskell/cabal/issues/9115, a new flag was discussed `--with-repl` which would render many of our rather hacky implementations superfluous.

Due to changes on how arguments are passed to `ghc` in `cabal`, cabal HEAD and the upcoming cabal 3.16 are currently completely incompatible with hie-bios and HLS.

To fix this, we update `hie-bios` to use `--with-repl` if the `cabal` version is new enough and the `lib:Cabal` version supports it.

fendor:
This avoids quite a lot of hacky implementations to find the correct GHC
version, libdir and ghc-pkg location, as well as global package db
stuff.

There is a corner case which we need to handle:

```
cabal >= 3.15 && lib:Cabal <3.15
```

In other words, when a custom-setup imposes a restriction on the
`lib:Cabal` version, and this `lib:Cabal` version doesn't support the
`--with-repl` flag, then we can't use `--with-repl`.

We detect this by looking at the error message of the cabal command, it
looks roughly like this:

    Error: [Cabal-7107]
    Could not resolve dependencies:
    [__0] trying: cabal-with-custom-setup-0.1.0.0 (user goal)
    [__1] next goal: cabal-with-custom-setup:setup.Cabal (dependency of cabal-with-custom-setup)
    [__1] rejecting: cabal-with-custom-setup:setup.Cabal; 3.10.3.0/installed-3.10.3.0, ... (constraint from --with-repl requires >=3.15)
    ...

We check the error message for the existence of `constraint from
--with-repl requires >=3.15` and then fall back to using the old
approach which doesn't use response files or `--with-repl` flag.

Additionally, we extract a new `Process` module which contains helpers
for cradle actions.

closes #467


src/HIE/Bios/Cradle.hs
```diff
@@ -762,32 +815,40 @@ processCabalWrapperArgs args =
...
-                      { cwd = Just wdir
-                      }
-          l <& LogCreateProcessRun ghcProc `WithSeverity` Debug
-          readCreateProcess ghcProc "" >>= putStr
-      else writeFile wrapper_fp wrapperContents
+withGhcWrapperTool l mkGhcCall wdir = do
+  withWrapperTool l mkGhcCall wdir "wrapper" cabalWrapperHs cabalWrapper
+
+-- | Generate a script/binary that can be passed to cabal's '--with-repl'.
+-- On windows, this compiles a Haskell file, while on other systems, we persist
```
mpickering: Incomplete comment.

src/HIE/Bios/Cradle.hs
```diff
@@ -909,70 +985,135 @@ cabalAction cradles workDir mc l projectFile fp loadStyle = do
...
-      | file <- fps,
-        -- Lookup the component for the old file
-        Just (ResolvedCradle {concreteCradle = ConcreteCabal ct, cradleDeps = deps}) <- [selectCradle prefix file (resolvedCradles cradles)],
-        -- Only include this file if the old component is in the same project
-        (projectConfigFromMaybe (cradleRoot cradles) (cabalProjectFile ct)) == projectFile,
-        let old_mc = cabalComponent ct
-      ]
+    -- | Run the given cabal process to obtain ghc options.
+    -- In the special case of 'cabal >= 3.15' but 'lib:Cabal <3.15' (via custom-setups),
+    -- we gracefully fall back to the given action to create an alterantive cabal process which
```
mpickering: `alternative`

mpickering (approved):
Looks fine to me in general, but I am not very familiar with this code and it changes many lines.

Is the approach load out somewhere in a Note? it would be good to precise explain at which cabal version this logic can be dropped.

PR #484, opened 2025-11-13
Fix processCabalLoadStyle to include dyn dep for extra files 

### LINKED ISSUE #481: Support correct dependency lookups for batch load mode
In batch load mod, we have the target file and and a list of files that might load with it.
However only the target file's dependencies' are correcttly setup. We need to find out how to setup the dependecies of the rest of the files correctly too.

soulomoon:
fix #481

The trick is to find the nearest cabal file for the target file inbetween itself and its cradleRoot

### DIFF
--- a/src/HIE/Bios/Cradle/Cabal.hs
+++ b/src/HIE/Bios/Cradle/Cabal.hs
@@ -20,6 +20,7 @@ import System.Exit
 import System.Directory
 import Colog.Core (LogAction (..), WithSeverity (..), Severity (..), (<&))
 import Control.Monad
+import Control.Monad.Extra (concatMapM)
 import Control.Monad.IO.Class
 import Data.Aeson ((.:))
 import qualified Data.Aeson as Aeson
@@ -35,7 +36,6 @@ import System.FilePath
 import System.Info.Extra (isWindows)
 import System.IO.Temp
 import Data.Version
-import Data.Tuple.Extra (fst3, snd3, thd3)
 
 import HIE.Bios.Config
 import HIE.Bios.Environment (getCacheDir)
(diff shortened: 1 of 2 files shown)


fendor:
Thanks for the fix! Feel free to merge once you are happy, or ping me if you want a second pair of eyes at some point!

Perhaps a testcase would be nice?

src/HIE/Bios/Cradle/Cabal.hs
```diff
@@ -332,8 +378,13 @@ processCabalWrapperArgs args =
...
 
 cabalLoadFilesBefore315 :: LogAction IO (WithSeverity Log) -> ProgramVersions -> CradleProjectConfig -> [Char] -> [String] -> CradleLoadResultT IO CreateProcess
-cabalLoadFilesBefore315 l progVersions projectFile workDir args = do
+cabalLoadFilesBefore315 l progVersions projectFile workDir args' = do
   let cabalCommand = "v2-repl"
+  cabal_version <- liftIO $ runCachedIO $ cabalVersion progVersions
+
+  let args = case cabal_version of
+        Just v | v < makeVersion [3,15] -> "--keep-temp-files" : args'
+        _ -> args'
```
soulomoon: For cabal version >= 3.15, we use cabalLoadFilesWithRepl with `--keep-temp-files` as one of the arg
For cabal version < 3.15 , we use cabalLoadFilesBefore315, where we do not have `--keep-temp-files`.

In windows, run without `--keep-temp-files`, `multi-out` file is being cleanup too early causing the problem.

Here I added back `--keep-temp-files` to keep `multi-out` around for cabalLoadFilesBefore315. Do you think it is a good workaround? @fendor

soulomoon: Yes, unfortunately somehow v2-repl does not take --keep-temp-files as argument after 3.15

fendor: Probably in conjunction with `--with-repl`, the flag isn't supported, but it looks like an oversight, tbh.

PR #487, opened 2025-11-26
Parallel test suite

alexfmpe:
On my M1 (`numCapabilities = 10`), without the stack tests:

Before `All 47 tests passed (79.10s)`

After: `All 47 tests passed (16.39s)`

### DIFF
--- a/tests/BiosTests.hs
+++ b/tests/BiosTests.hs
@@ -15,16 +15,19 @@ import qualified Test.Tasty.Options as Tasty
 import qualified Test.Tasty.Ingredients as Tasty
 import HIE.Bios
 import HIE.Bios.Cradle
-import Control.Monad ( forM_ )
+import Control.Monad (forM_)
+import Control.Monad.Extra (unlessM)
+import Control.Monad.IO.Class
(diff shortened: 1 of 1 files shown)


src/HIE/Bios/Environment.hs
```diff
@@ -28,12 +28,23 @@ import qualified HIE.Bios.Ghc.Gap as Gap
...
     -- Create a unique folder per set of different GHC options, assuming that each different set of
     -- GHC options will create incompatible interface files.
-    let opts_hash = B.unpack $ encode $ H.finalize $ H.updates H.init (map B.pack componentOptions)
+    let
+      -- There seems to be a race condition when writing interface files
+      hash_args = (if workAroundThreadUnsafety then (componentRoot :) else id) componentOptions
+      opts_hash = B.unpack $ encode $ H.finalize $ H.updates H.init $ map B.pack hash_args
+
     cache_dir <- liftIO $ getCacheDir opts_hash
+    liftIO $ createDirectoryIfMissing True cache_dir
```
fendor: Is this necessary?
Why would this change only be necessary now?

Issue #485, opened 2025-11-17
`cabal path` fails with cabal HEAD


fendor:
If you use `cabal` HEAD, or any local install, often you use a debug installation of cabal. In this case, `cabal` will output the warning:

```
Warning: this is a debug build of cabal-install with assertions enabled.
```

for each cli invocation.
This output warning can't be silenced.

With this output, the output of `cabal path` is unexpectedly not a valid json object.
Thus, `hie-bios` can't find the ghc options and renders HLS inoperable with cabal HEAD installations that have assertions enabled.

The solution is to invoke `cabal path --output-format=json -vnormal+markoutput` and parse the last output block:

```sh
> cabal path --output-format=json -vnormal+markoutput
-----BEGIN CABAL OUTPUT-----
Configuration is affected by the following files:
- cabal.project
-----END CABAL OUTPUT-----
-----BEGIN CABAL OUTPUT-----
{"cabal-version":"3.14.2.0","compiler":{"flavour":"ghc","id":"ghc-9.10.3","path":"/home/hugin/.ghcup/bin/ghc"},"cache-home":"/home/hugin/.cache/cabal","remote-repo-cache":"/home/hugin/.cache/cabal/packages","logs-dir":"/home/hugin/.cache/cabal/logs","store-dir":"/home/hugin/.local/state/cabal/store","config-file":"/home/hugin/.config/cabal/config","installdir":"/home/hugin/.local/bin"}
-----END CABAL OUTPUT-----
```

PR #499, opened 2026-02-27
Handle noisy stdout from cabal (e.g. darcs warnings)

### LINKED ISSUE #498: GHC detection fails in a Darcs-managed project that depends on another package via source-repository-package with type: darcs
When you use `darcs clone` to clone a repo inside an existing darcs repo, it outputs `WARNING: creating a nested repository.`. This seems to interfere with parsing out the GHC information. The upshot is that if you have a `cabal.project` in a darcs-managed project that specifies that some dependency comes from a darcs repository, HLS fails.

jmcarthur:
Fixes #498.

When a cabal project has a `source-repository-package` with `type: darcs`, cabal triggers `darcs clone` during various operations. darcs likes to print `WARNING: creating a nested repository.` to stdout, and this noise ends up in the output of several commands that hie-bios runs and parses. The result is that HLS fails to start entirely in darcs-managed projects with darcs dependencies.

There were three places that needed fixing:

`versionMaybe` in ProgramVersions.hs was parsing the raw output of `--numeric-version` without stripping leading noise. The `trim` function (take the last line, strip trailing whitespace) already existed in Environment.hs for exactly this kind of problem; the comment even mentions Stack's 7zip noise. I moved `trim` to `Cradle.Utils` so both modules can use it.

`cabalGhcDirs` and `cabalProcess` in Cabal.hs were using `trimEnd` (strip trailing whitespace only) on the output of `cabal exec -- ghc --print-libdir` and `cabal exec -- ghc -e getExecutablePath`. This meant `HIE_BIOS_GHC` and `HIE_BIOS_GHC_ARGS` got the darcs warnings prepended to the actual paths. Changed these to use `trim` as well.

`callCabalPathForCompilerPath` parses JSON from `cabal path --output-format=json`, but the darcs warnings appear before the JSON object. I added a `dropWhile` to skip to the first `{` before parsing. Without this fix it technically still works (it falls back to the slower `cabal exec` path) but it does so on every single file load, making startup very slow.

Having to apply essentially the same fix in so many places feels like a code smell. There might be an opportunity to centralize the output cleanup closer to the process-running layer so that individual call sites don't each have to remember to strip noise.

I added an integration test for the `versionMaybe` path using a bios cradle with a noisy GHC wrapper script. The Cabal.hs fixes are harder to test without darcs in CI, but they use the same `trim` function. I tested the whole thing end-to-end against a real darcs project and HLS starts up and type-checks everything. Open to suggestions on better test coverage.

### DIFF
--- a/hie-bios.cabal
+++ b/hie-bios.cabal
@@ -145,6 +145,11 @@ Extra-Source-Files:     README.md
                         tests/projects/failing-multi-repl-cabal-project/multi-repl-cabal-fail/src/Fail.hs
                         tests/projects/failing-multi-repl-cabal-project/multi-repl-cabal-fail/src/Lib.hs
                         tests/projects/failing-multi-repl-cabal-project/NotInPath.hs
+                        tests/projects/noisy-bios-ghc/A.hs
+                        tests/projects/noisy-bios-ghc/B.hs
+                        tests/projects/noisy-bios-ghc/hie-bios.sh
+                        tests/projects/noisy-bios-ghc/noisy-ghc.sh
+                        tests/projects/noisy-bios-ghc/hie.yaml
 
 tested-with: GHC ==9.6.7 || ==9.8.4 || ==9.10.1 || ==9.12.2 || ==9.14.1
 
--- a/src/HIE/Bios/Cradle/Cabal.hs
+++ b/src/HIE/Bios/Cradle/Cabal.hs
@@ -403,7 +403,7 @@ cabalProcess l vs cabalProject workDir command args = do
   ghcDirs@(ghcBin, libdir) <- callCabalPathForCompilerPath l vs workDir cabalProject >>= \case
     Just p -> do
       libdir <- Process.readProcessWithCwd_ l workDir p ["--print-libdir"] ""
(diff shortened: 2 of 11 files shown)


fendor:
Have you seen #485? The issue describes a slightly better way to parse the output without semi-random `trim`ming and `/= '{'`.

Perhaps this could be a good idea?

jmcarthur:
I think you're right. I won't look at it for at least a few days (TBH, this isn't super high priority for me right now), but I'll get around to it.

jmcarthur:
In fact, I've realized since fixing this bug for myself that, at least for my private projects, I don't think I want to depend on packages via source control, because it causes the build to check upstream every time (I'm not pinning the versions since it's private and I control the dependency anyway), so I'm probably going to just go back to relative, local paths. So this bug doesn't actually matter to me anymore. I'll still take a stab at doing it the better way, maybe within a week, but if it doesn't work out easily, I'll probably abandon it.

fendor:
Sounds sensible, thank you for raising this issue and providing an idea to fix it!

Hope to see you back in a couple of weeks ;)

Issue #501, opened 2026-03-15
When using `--with-repl`, `--builddir` is not passed.


mniip:
hie-bios (and by proxy HLS) will use `cabal repl --with-repl` when cabal is of version `>=3.15`. However in this branch `--builddir` is not passed to `cabal`.

In combination with https://github.com/haskell/cabal/issues/11623 this clobbers the main build directory and makes subsequent builds with `cabal build` slower than they need to be.

Using `hie-bios debug Test.hs` directly, we see:
```
[Info] cabal v2-repl --keep-temp-files --with-repl /home/mniip/.cache/hie-bios/repl-wrapper-fae20739a232ed1b9dcdf40eca9cf71e Test.hs
  Environment Variables
    HIE_BIOS_OUTPUT: /tmp/HIE_BIOS_OUTPUT3665214-0
```
where with an older cabal version, we see:
```
[Info] cabal --builddir=/home/mniip/.cache/hie-bios/dist-test-cabal-project-29747344345f3196b93f25ef770266a8 v2-repl --with-compiler /home/mniip/.cache/hie-bios/wrapper-b54f81dea4c0e6d1626911c526bc4e36 --with-hc-pkg /home/mniip/.cache/hie-bios/ghc-pkg-296587c7824bd05edf701243e6027224 Test.hs
  Environment Variables
    HIE_BIOS_OUTPUT: /tmp/HIE_BIOS_OUTPUT3656552-0
    HIE_BIOS_GHC: /home/mniip/.ghcup/bin/ghc
    HIE_BIOS_GHC_ARGS: -B/home/mniip/.ghcup/ghc/9.8.4/lib/ghc-9.8.4/lib
```

Bisecting on presence of `--builddir`, we see that 4ac7208d112955dbb7f26b48f165043ec52748eb (#466) is the first bad commit.

@fendor mentioned
>  [2:24 PM] fendor: I am not 100% sure, but we might not show in the logs that we are providing a custom --builddir... At least, I think we should be using a custom builddir

I've verified the lack of `--builddir` in `/proc/$pid/cmdline`.

sheaf:
Oh this bug has been driving me crazy. I put up #503 which should fix it.

PR #503, opened 2026-03-31
Pass --builddir to avoid clobbering dist-newstyle

sheaf:
Ticket: #501 

This PR ensures we always pass `--builddir` to cabal-install, in particular when using `--with-repl`. This is important to avoid HLS sessions trampling over the user's local dist-newstyle directory, causing unnecessary recompilation.

I have checked that the added test fails without the change in this PR.

### DIFF
--- a/src/HIE/Bios/Cradle/Cabal.hs
+++ b/src/HIE/Bios/Cradle/Cabal.hs
@@ -11,9 +11,11 @@ module HIE.Bios.Cradle.Cabal
   withGhcWrapperTool,
   -- * Argument processing
   processCabalWrapperArgs,
-  -- * Exposed for tests
+  -- * Internals (exposed for tests)
   isCabalMultipleCompSupported,
-  )where
+  cabalBuildDir,
+  )
+  where
 
 import Data.Char (isSpace)
 import System.Exit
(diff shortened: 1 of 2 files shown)


fendor (approved):
LGTM, thank you!

sheaf:
I will try to investigate the Windows failures.

However, when using HLS on GHC I'm still seeing "Configuration changed" and unnecessary rebuilds, so perhaps there is still something else going on? This might just be a Hadrian bug however.

fendor:
> This might just be a Hadrian bug however.

It most likely is

> I will try to investigate the Windows failures.

I am wondering whether this is caused by the parallel test execution and windows's handling of concurrent file access. Perhaps running the windows tests without parallelism suffices.

fendor:
> However, when using HLS on GHC I'm still seeing "Configuration changed" and unnecessary rebuilds

Just to emphasize, HLS on GHC doesn't use the cabal cradle, so this fix should be unrelated.

sheaf:
> > However, when using HLS on GHC I'm still seeing "Configuration changed" and unnecessary rebuilds
> 
> Just to emphasize, HLS on GHC doesn't use the cabal cradle, so this fix should be unrelated.

Yes, after a bit more investigation I think the GHC issues were user error on my part (combined with some Hadrian bugs about not invalidating the cache when going from a release build to a non-release build).

PR #508, opened 2026-06-10
[Fix #507] New Load Modes for upfront loading of cradle/project

### LINKED ISSUE #507: New Load Modes for upfront loading of cradle/project
## Motivation

hie-bios users like HDB want to load the cradle to discover files, so the existing `LoadContext` interface is inadequate for projects with multiple units.
See: https://github.com/well-typed/haskell-debugger/issues/38

## API

```haskell
type Component = String

-- | Always needed for graceful fallback to (1.1) and (1.2)
data TargetWithContext = TargetWithContext
    { targetFilePath :: FilePath
    , targetContext :: [FilePath]
    }

data CradleLoadMode
    = LoadFile
    | LoadFileTargetWithContext
    | LoadUnitsInferred
    | LoadUnitsFromCradle

-- | Note: @getCompilerOptions (TargetWithContext path cxt) LoadFile@ will warn if `cxt` is non-empty.
getCompilerOptions :: TargetWithContext -> CradleLoadMode -> IO ComponentOptions
```

## Spec

We ignore details such as logging, working directory and error handling.

This design mostly affects cabal and perhaps the bios cradle. Stack cradle is unaffected right now.

1. **`LoadFile`**, **`LoadFileTargetWithContext`**: These are meant to match the behaviour of the previous API (i.e. `LoadStyle`): load only components that match the provided file (and context).
    * More specifically:
        1. `LoadFile`: ignore context
            * simplest interface, cradle impls should avoid this mode
            * default cradle will load just this single file.
            * Essentially equivalent to: `Maybe Component -> FilePath -> IO ComponentOptions`
              * `Maybe Component` is the relevant info from the cradle
        2. `LoadFileTargetWithContext`: uses context
            * Loads components matching the files given by `paths = targetFilePath : targetContext`.
            * Cabal invocation: `cabal repl --enable-multi-repl`
            * Not supported by: stack, bios, direct (direct is always assumed to be a single module)
            * Can gracefully fall back to `Legacy Mode` if not supported by cradle
2. **`LoadUnitsFromCradle`**: loads all units that match any of the components specified in the crade.
    * Equivalent to `[Component] -> IO ComponentOptions`
      * `[Component]` is from the cradle
    * Cabal invocation: `cabal repl --enable-multi-repl component1 component2 ... componentN`
    * Fallback:
        * Falls back to (3) if mix of components in cradle does not support being loaded at once.
        * Falls back to (1.2) when (3) is not supported either.
3. **`LoadUnitsInferred`**: loads all units from current project, whether they have a cradle component or not.
    * Equivalent to `IO ComponentOptions`
    * Respects project configuration (but see issue in examples)
    * Fallback:
        * Falls back to (1.2) when not supported.

These also generally apply:
  * Any dependencies of the components selected for loading, are pre-installed. This applies also to dependencies that are units from the current project.
  * The `TargetWithContext` argument of `getCompilerOptions` is passed along during fallback, so it's relevant for all modes.

The above means there are still 4 different ways of loading a project, where only `cabal` Cradles currently support all session initialisation things.
`bios` could be easily thaught about it as well, but `stack` is very tricky.

### `LoadUnitsInferred` examples
Given no hie.yaml, and `LoadUnitsInferred`, this would be the equivalent of:

```
cabal repl --enable-multi-repl all <filepath> --enable-tests --enable-benchmarks
```

So that most modules in the project are loaded. 

Note that the user might have specified the opposite in the `cabal config` or `cabal.project[.local]` files, so this behavior might be surprising. Unfortunately cabal's interface doesn't have a way to tell if such explicit configurations are active nor to specify that the given flags should not override explicit configs.

If we do have a `hie.yml` that specifies a cabal project then we just follow that
rather than add further flags (see example below). Users wishing to avoid --enable-test and/or --enable-benchmarks can rely on this.

---

-- hie.yaml
```yaml
cradle:
  cabal:
    cabalProject: "./cabal.project.dev"
```

```
cabal --project-file ./cabal.project.dev repl --enable-multi-repl all <filepath>
```

### `LoadUnitsFromCradle` examples
---

-- hie.yaml
```yaml
cradle:
  cabal:
    cabalProject: "./cabal.project.dev"
    components:
    - path: "./src"
      component: "lib:hie-bios"
    - path: "./exe"
      component: "exe:hie-bios"
    - path: "./tests/BiosTests.hs"
      component: "test:hie-bios"
    - path: "./tests/ParserTests.hs"
      component: "test:parser-tests"
```

```
cabal repl --enable-multi-repl lib:hie-bios exe:hie-bios test:hie-bios test:parser-tests
```

---

-- hie.yaml
```yaml
cradle:
  cabal:
    component: "lib:hie-bios"
```

```
cabal repl lib:hie-bios
```

Saizan:
PR for #507

### DIFF
--- a/src/HIE/Bios/Cradle.hs
+++ b/src/HIE/Bios/Cradle.hs
@@ -139,7 +139,6 @@ resolvedCradlesToCradle logger buildCustomCradle root cs = mdo
     , cradleOptsProg = CradleAction
       { actionName = multiActionName
       , runCradle  = \fpc prev -> do
-          -- TODO: LoadUnits* ?
           absfp <- makeAbsolute (targetFilePath fpc)
           case selectCradle (prefix . fst) absfp cradleActions of
             Just (rc, act) -> do
--- a/src/HIE/Bios/Cradle.hs
+++ b/src/HIE/Bios/Cradle.hs
@@ -94,11 +94,13 @@ loadImplicitCradle l wfile = do
 loadCradleWithOpts :: (Yaml.FromJSON b, Show a) => LogAction IO (WithSeverity Log) -> (b -> CradleAction a) -> FilePath -> IO (Cradle a)
 loadCradleWithOpts l buildCustomCradle wfile = do
     cradleConfig <- readCradleConfig wfile
+    l <& WithSeverity (LogAny $ T.pack $ "Cradle Config: " ++ show cradleConfig) Debug
     getCradle l buildCustomCradle (cradleConfig, takeDirectory wfile)
 
 getCradle :: Show a => LogAction IO (WithSeverity Log) ->  (b -> CradleAction a) -> (CradleConfig b, FilePath) -> IO (Cradle a)
 getCradle l buildCustomCradle (cc, wdir) = do
     rcs <- canonicalizeResolvedCradles wdir cs
+    liftIO $ l <& WithSeverity (LogAny . T.pack $ "Resolved Cradles " ++ show ((fmap . fmap) (const ()) rcs)) Debug
     resolvedCradlesToCradle l buildCustomCradle wdir rcs
   where
     cs = resolveCradleTree wdir cc
--- a/src/HIE/Bios/Cradle.hs
+++ b/src/HIE/Bios/Cradle.hs
@@ -381,13 +381,20 @@ biosCradle l rc wdir biosCall biosDepsCall mbGhc
 biosWorkDir :: FilePath -> MaybeT IO FilePath
 biosWorkDir = Process.findFileUpwards ".hie-bios"
 
+-- | shared between @biosDepsAction@ and @biosAction@.
+biosFilePaths :: TargetWithContext -> LoadMode -> [FilePath]
+biosFilePaths fpc = \case
+  LoadFile -> [targetFilePath fpc]
+  LoadFileWithContext -> targetAndContext fpc
+  -- Exhaustive matching to trigger type errors for datatype changes.
+  LoadUnitsInferred -> targetAndContext fpc
+  LoadUnitsFromCradle -> targetAndContext fpc
+
 biosDepsAction :: LogAction IO (WithSeverity Log) -> FilePath -> Maybe Callable -> TargetWithContext -> LoadMode -> IO [FilePath]
 biosDepsAction l wdir (Just biosDepsCall) fpc loadStyle = do
-  let fp = targetFilePath fpc
 
-  let fps = case loadStyle of
-        LoadFile -> [fp]
-        LoadFileWithContext -> fp : targetContext fpc
+  let fps = biosFilePaths fpc loadStyle
+
   (ex, sout, serr, [(_, args)]) <-
     runContT (withCallableToProcess biosDepsCall fps) $ \biosDeps' ->
       Process.readProcessWithOutputs [hie_bios_output] l wdir biosDeps'
@@ -408,32 +415,37 @@ biosAction
 biosAction rc wdir bios bios_deps l fpc loadStyle = do
   let fp = targetFilePath fpc
   ghc_version <- liftIO $ runCachedIO $ ghcVersion $ cradleProgramVersions rc
+  let warnUnsupportedLoadMode msg =
+        liftIO $ l <& WithSeverity
+          (LogLoadModeUnsupported "bios" loadStyle (Just msg))
+          Warning
   determinedLoadMode <- case ghc_version of
-    Just ghc
-      -- Multi-component supported from ghc 9.4
-      -- We trust the assertion for a bios program, as we have no way of
-      -- checking its version
-      | LoadFileWithContext <- loadStyle ->
-          if ghc >= makeVersion [9,4]
-            then pure loadStyle
-            else do
-              liftIO $ l <& WithSeverity
-                (LogLoadFileWithContextUnsupported "bios"
-                  $ Just "ghc version is too old. We require `ghc >= 9.4`"
-                )
-                Warning
-              pure LoadFile
-    _ -> pure LoadFile
-  let fps = case determinedLoadMode of
-        LoadFile -> [fp]
-        LoadFileWithContext -> fp : targetContext fpc
+      Just ghc
+        | ghc >= makeVersion [9,4] -> do
+          -- Multi-component supported from ghc 9.4
+          -- We trust the assertion for a bios program, as we have no way of
+          -- checking its version
+          pure $ case loadStyle of
+            LoadFile -> LoadFile
+            LoadFileWithContext -> LoadFileWithContext
+            LoadUnitsFromCradle -> LoadFileWithContext
+            LoadUnitsInferred   -> LoadFileWithContext
+        | loadStyle /= LoadFile -> do
+          warnUnsupportedLoadMode "ghc version is too old. We require `ghc >= 9.4`"
+          pure LoadFile
+      _ -> do
+        warnUnsupportedLoadMode "Unable to determine ghc version. We require `ghc >= 9.4`"
+        pure LoadFile
+
+  let fps = biosFilePaths fpc determinedLoadMode
+
   (ex, _stdo, std, [(_, res),(_, mb_deps)]) <-
     runContT (withCallableToProcess bios fps) $ \bios' ->
       Process.readProcessWithOutputs [hie_bios_output, hie_bios_deps] l wdir bios'
 
   deps <- case mb_deps of
     Just x  -> return x
-    Nothing -> biosDepsAction l wdir bios_deps fpc loadStyle
+    Nothing -> biosDepsAction l wdir bios_deps fpc determinedLoadMode
         -- Output from the program should be written to the output file and
         -- delimited by newlines.
         -- Execute the bios action and add dependencies of the cradle.
@@ -690,9 +702,9 @@ runGhcCmdOnPath l wdir args = Process.readProcessWithCwd l wdir "ghc" args ""
 -- | Log that the cradle has no supported for loading with context, if and only if
 -- 'LoadFileWithContext' was requested.
 logCradleHasNoSupportForLoadFileWithContext :: Applicative m => LogAction m (WithSeverity Log) -> LoadMode -> T.Text -> m ()
-logCradleHasNoSupportForLoadFileWithContext l (LoadFileWithContext) crdlName =
+logCradleHasNoSupportForLoadFileWithContext l LoadFileWithContext crdlName =
   l <& WithSeverity
-        (LogLoadFileWithContextUnsupported crdlName
+        (LogLoadModeUnsupported crdlName LoadFileWithContext
           $ Just $ crdlName <> " doesn't support loading multiple components at once"
         )
         Info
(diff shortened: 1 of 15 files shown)


src/HIE/Bios/Environment.hs
```diff
@@ -159,7 +159,7 @@ addCmdOpts cmdOpts df1 = do
   additionalTargets <- concat <$> mapM (liftIO . getTargetsFromGhciScript) (G.ghciScripts df2)
 
   -- leftovers contains all Targets from the command line
-  let leftovers = map G.unLoc leftovers' ++ additionalTargets
+  let leftovers = filter (/= "-unit") (map G.unLoc leftovers') ++ additionalTargets
```
Saizan: @fendor does it make sense that I'd have to filter "-unit" here? Or am I doing something wrong?

fendor: I don't know why you should have to filter the `-unit` flags out here, that looks incorrect.

We didn't have to do this before, iirc.

Saizan: Turns out that initSession/addCmdOpts are not setup to handle multiple home units, i guess we should update them? are they used in hls?

fendor: No, I think we are using them only for tests these days.

EDIT: addCmdOpts is still used in HLS for initialising a single home unit.

Saizan: I left addCmdOpts alone. initSession' now uses addCmdOpts if there is no `-unit` flag, and initMulti otherwise

src/HIE/Bios/Ghc/Gap.hs
```diff
@@ -303,3 +323,143 @@ unsafeEncodeUtf fp =
...
+#else
+
+-- -- Strip out any ["+RTS", ..., "-RTS"] sequences in the command string list.
+-- removeRTS :: [String] -> [String]
+-- removeRTS ("+RTS" : xs)  =
+--   case dropWhile (/= "-RTS") xs of
+--     [] -> []
+--     (_ : ys) -> removeRTS ys
+-- removeRTS (y:ys)         = y : removeRTS ys
+-- removeRTS []             = []
```
fendor:
```suggestion
```

src/HIE/Bios/Cradle/Cabal.hs
```diff
@@ -296,6 +307,33 @@ processCabalLoadStyle l cradles projectFile workDir mc fp loadStyle = do
...
+            |rc@(ResolvedCradle {concreteCradle = ConcreteCabal ct, cradleDeps = depsYaml}) <- (resolvedCradles cradles)
+            , (projectConfigFromMaybe (cradleRoot cradles) (cabalProjectFile ct)) == projectFile
+            , Just comp <- [cabalComponent ct]
+            ]
+      -- TODO: warn about presence of other components, unless of type None.
+      -- TODO: should calc deps from prefixes here instead?
+      pure [comp | (_,_,comp) <- selected]
+    loadUnits explicitComps = do
+      let fps = targetContext fpc
+      (modPairs, mergedDeps) <- moduleFilesFromSameProject fps
```
Saizan: double check it covers all the deps of all the components

Saizan:
@fendor I punted the `step` and yaml format improvements to issues and CI is happy, shall we merge?

fendor:
@Saizan I'll proofread it one more time, then yes!

src/HIE/Bios/Ghc/Gap.hs
```diff
@@ -303,3 +323,135 @@ unsafeEncodeUtf fp =
...
+  initial_dflags <- GHC.getSessionDynFlags
+
+  dynFlagsAndSrcs <- forM unitArgsFiles $ \f -> do
+    when (verbosity initial_dflags > 2) (liftIO $ print f)
+    args <- liftIO $ expandResponse [f]
+    (dflags2, fileish_args, _warns) <- parseDynamicFlagsCmdLine initial_dflags (map (mkGeneralLocated f) (removeRTS args))
+    -- handleSourceError (\e -> do
+    --    GHC.printException e
+    --    liftIO $ exitWith (ExitFailure 1)) $ do
+    --      liftIO $ handleFlagWarnings logger (initPrintConfig dflags2) (initDiagOpts dflags2) warns
```
fendor:
```suggestion
```

Delete?

src/HIE/Bios/Ghc/Gap.hs
```diff
@@ -303,3 +323,135 @@ unsafeEncodeUtf fp =
...
+
+    let (dflags3, srcs, _objs) = parseTargetFiles dflags2 (map unLoc fileish_args)
+        dflags4 = offsetDynFlags dflags3
+
+    let (hs_srcs, _non_hs_srcs) = partition isHaskellishTarget srcs
+
+
+
```
fendor:
```suggestion
```

src/HIE/Bios/Ghc/Gap.hs
fendor: Or are you intentionally keeping this in to show how it looks like in GHC?

src/HIE/Bios/Cradle.hs
```diff
@@ -379,11 +382,21 @@ biosCradle l rc wdir biosCall biosDepsCall mbGhc
 biosWorkDir :: FilePath -> MaybeT IO FilePath
 biosWorkDir = Process.findFileUpwards ".hie-bios"
 
-biosDepsAction :: LogAction IO (WithSeverity Log) -> FilePath -> Maybe Callable -> FilePath -> LoadStyle -> IO [FilePath]
-biosDepsAction l wdir (Just biosDepsCall) fp loadStyle = do
-  let fps = case loadStyle of
-        LoadFile -> [fp]
-        LoadWithContext old_fps -> fp : old_fps
+-- | shared between @biosDepsAction@ and @biosAction@.
+--   TODO: should biosDepsAction also fallback to LoadFile for ghc < 9.14 ?
```
fendor:
```suggestion
```

We do?

src/HIE/Bios/Types.hs
```diff
@@ -156,29 +167,46 @@ instance Pretty Log where
     "Could not parse json output of 'cabal path': "
       <> line <> indent 4 (pretty err)
 
--- | The 'LoadStyle' instructs a cradle on how to load a given file target.
-data LoadStyle
+type Component = String
+
+-- | Always needed for graceful fallback to (1.1) and (1.2)
```
fendor: This is referring to some docs I can't see here, perhaps we should outline the design here?

tests/BiosTests.hs
```diff
@@ -461,6 +503,18 @@ biosTestCase name assertion = askOption @VerboseLogging (\case
...
+biosTestCaseMulti = biosTestCaseAll' [LoadFileWithContext .. ] testGroup
+
+biosTestCaseAll :: TestName -> (LoadMode -> Bool -> Assertion) -> TestTree
+biosTestCaseAll = biosTestCaseAll' [minBound .. ] testGroup
+
+biosTestCaseAllInOrder :: TestName -> (LoadMode -> Bool -> Assertion) -> TestTree
+biosTestCaseAllInOrder = biosTestCaseAll' [minBound .. ] inOrderTestGroup
+
+biosTestCaseAll' :: [LoadMode] -> (String -> [TestTree] -> TestTree) -> TestName -> (LoadMode -> Bool -> Assertion) -> TestTree
+biosTestCaseAll' modes tG name assertion = tG name $ [biosTestCase (name ++ ":" ++ show mode) (assertion mode) | mode <- modes ]
```
fendor: I think ":" may be a special character, and we need to escape when trying to rerun a test, e.g.

```
> TASTY_PATTERN="failing-cabal:LoadFile" cabal test
parser-tests: Bad environment variable TASTY_PATTERN='failing-cabal:LoadFile' (parsed as option pattern)
HasCallStack backtrace:
  collectBacktraces, called at libraries/ghc-internal/src/GHC/Internal/Exception.hs:169:13 in ghc-internal:GHC.Internal.Exception
  toExceptionWithBacktrace, called at libraries/ghc-internal/src/GHC/Internal/IO.hs:260:11 in ghc-internal:GHC.Internal.IO
  throwIO, called at ./Test/Tasty/Options/Env.hs:49:17 in tasty-1.5.4-22543be7f26a82be3cfaa0786f06221ba556f63aa7729e556dd906a4821c8fee:Test.Tasty.Options.Env
```

You have to invoke it as `TASTY_PATTERN="/failing-cabal:LoadFile/" cabal test`. Perhaps not a big loss, but something to keep in mind.

tests/BiosTests.hs
```diff
@@ -172,13 +175,13 @@ biosTestCases =
 cabalTestCases :: ToolDependency -> [TestTree]
 cabalTestCases extraGhcDep =
   [
-    biosTestCase "failing-cabal" $ runTestEnv "./failing-cabal" $ do
-      cabalAttemptLoad "MyLib.hs"
+    biosTestCaseAll "failing-cabal" $ \mode -> runTestEnv "./failing-cabal" $ do
+      cabalAttemptLoadWithMode' mode "MyLib.hs"
```
fendor, line 179: How about moving the `LoadMode` to `TestEnv`? Only the leafs of the test helpers would need to check it, and it would avoid having to pass `mode` everywhere.

We can also leave it as it is and change it later on.

fendor (approved):
LGTM, a couple of nitpicks, feel free to ignore.

Nice patch, thank you very much!

src/HIE/Bios/Ghc/Gap.hs
```diff
@@ -303,3 +323,135 @@ unsafeEncodeUtf fp =
...
+  -- if we have no haskell sources from which to do a dependency
+  -- analysis, then just do one-shot compilation and/or linking.
+  -- This means that "ghc Foo.o Bar.o -o baz" links the program as
+  -- -- we expect.
+  -- if (null hs_srcs)
+  --     then do
+  --       liftIO $ hPutStrLn stderr $ "Multi Mode can not be used for one-shot mode."
+  --       liftIO $ exitWith (ExitFailure 1)
+  --     else do
```
Saizan: remove this too

src/HIE/Bios/Cradle.hs
```diff
@@ -398,37 +411,43 @@ biosAction
...
+  let fps = biosFilePaths fpc determinedLoadMode
+
   (ex, _stdo, std, [(_, res),(_, mb_deps)]) <-
     runContT (withCallableToProcess bios fps) $ \bios' ->
       Process.readProcessWithOutputs [hie_bios_output, hie_bios_deps] l wdir bios'
 
   deps <- case mb_deps of
     Just x  -> return x
-    Nothing -> biosDepsAction l wdir bios_deps fp loadStyle
+    Nothing -> biosDepsAction l wdir bios_deps fpc loadStyle
```
Saizan: should be determinedLoadMode

src/HIE/Bios/Cradle.hs
Saizan: We don't yet because we call it with `loadStyle` rather than `determinedLoadMode`, but I made a comment about it below

tests/BiosTests.hs
Saizan: Let's do it as part of some other testsuite refactor.

tests/BiosTests.hs
Saizan: Fair, I'll change it to `-`.

fendor (approved):

PR #513, opened 2026-06-22
[Fix #512] Fail hard if component(s) field have wrong type

### LINKED ISSUE #512: "components" field silently ignored when not a list.
A cradle like the following is parsed as if no components were given, because "components" expects a list and the parser just falls back to no components when parsing "exe:t3" as a list fails.
```
 cradle:
   cabal:
    components: "exe:t3"
```

Saizan:
Once we recognize the presence of the field we commit to parsing it.


fendor (approved):
LGTM, thank you!

PR #515, opened 2026-06-24
[Fix #517] "componentsToLoad" field

Saizan:
@fendor added the field for StackConfig too, but not sure about it because we don't really have a way to implement it right?

### DIFF
--- a/README.md
+++ b/README.md
@@ -554,28 +554,29 @@ cradle:
 
 ## Selecting which components to load
 
-For `cabal` and `stack` cradles, tools using the `LoadUnitsFromCradle` load mode
-will load all the components listed, ignoring paths.
+For `cabal` and `stack` cradles, `hie-bios` supports asking for flags to load
+the components in the cradle at once. This is much faster than loading
+components on-demand at the cost of more memory usage if a component isn't
+actually needed. Support is conditional on underlying tool versions, and
+`hie-bios` will fall back to legacy loading otherwise.
 
-The additional field `componentsToLoad` can be used to override the list of components loaded via `LoadUnitsFromCradle`.
+The optional field `componentsToLoad` specifies which components should be loaded together, it defaults to all the explicitly listed components.
 
 ```yaml
 cradle:
-  config:
-    cabal:
-      componentsToLoad: ["lib:foo"]
-      components:
-        - {"path": "./src", "component": "lib:foo"}
-        - {"path": "./app", "component": "exe:foo"}
+  cabal:
+    componentsToLoad: ["lib:foo"]
+    components:
+      - {"path": "./src", "component": "lib:foo"}
+      - {"path": "./app", "component": "exe:foo"}
 ```
 
 The elements of `componentsToLoad` do not need to be also listed as individual components, though doing so might improve dependency tracking.
 
 ```yaml
 cradle:
-  config:
-    stack:
-      componentsToLoad: ["foo:lib","foo:exe"]
+  stack:
+    componentsToLoad: ["foo:lib","foo:exe"]
 ```
 
 ## Cradle Dependencies
@@ -636,6 +637,7 @@ The complete configuration is a subset of
 cradle:
   cabal:
     cabalProject: "optional path to project description"
+    componentsToLoad: ["cabal target1","cabal target2"]
     -- one or none of the following
     component: "optional component name"
     components:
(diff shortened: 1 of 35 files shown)


fendor:
I think `stack repl comp1 comp2` works as long as only 1 executable is listed... Otherwise `stack repl` will ask for which component you want to treat as the main executable and never start up.

I am not sure whether that's a big footgun, but it honestly sounds like exactly what we need to support `stack` better in the haskell-debugger

Saizan:
I see, I'll implement the feature for stack then

fendor (approved):
LGTM, could we also document the new fields in the README?

A disadvantage of `componentsToLoad` is that it is difficult to find the `.cabal` cradle dependencies for automatic reload 🤔

README.md
```diff
@@ -609,9 +635,19 @@ The complete configuration is a subset of
 ```yaml
 cradle:
   cabal:
+    cabalProject: "optional path to project description"
+    -- one or none of the following
     component: "optional component name"
+    components:
```
fendor, line 643:
```suggestion
    componentsToLoad: ["cabal target"]
    components:
```

README.md
```diff
@@ -609,9 +635,19 @@ The complete configuration is a subset of
...
+      - path: "path to package dir"
+        component: "cabal target"
   stack:
+    stackYaml: "optional path to stack yaml file"
+    -- one or none of the following
     component: "optional component name"
+    components:
```
fendor, line 651:
```suggestion
    componentsToLoad: ["stack target"]
    components:
```

README.md
```diff
@@ -552,6 +552,32 @@ cradle:
                             , { path: "./tests", component: "parser-tests" } ] } }
 ```
 
+## Selecting which components to load
+
+For `cabal` and `stack` cradles, tools using the `LoadUnitsFromCradle` load mode
+will load all the components listed, ignoring paths.
+
+The additional field `componentsToLoad` can be used to override the list of components loaded via `LoadUnitsFromCradle`.
```
fendor: I think we shouldn't mention the load mode, it is not mentioned anywhere else and without context doesn't tell the user anything.

I would rather say something like:

```suggestion
For `cabal` and `stack` cradles, we can list all components that should be part of the GHC session and load them all at once. This is much faster than loading them on-demand at the cost of more memory usage if a component isn't actually needed.

The optional field `componentsToLoad` specifies which components should be loaded together if the cradle supports it. If the cradle doesn't support it, it will fall back to legacy session loading.
```

README.md
```diff
@@ -552,6 +552,32 @@ cradle:
...
+
+```yaml
+cradle:
+  config:
```
fendor:
```suggestion
```

README.md
```diff
@@ -552,6 +552,32 @@ cradle:
...
+      components:
+        - {"path": "./src", "component": "lib:foo"}
+        - {"path": "./app", "component": "exe:foo"}
+```
+
+The elements of `componentsToLoad` do not need to be also listed as individual components, though doing so might improve dependency tracking.
+
+```yaml
+cradle:
+  config:
```
fendor:
```suggestion
```

README.md
```diff
@@ -552,6 +552,32 @@ cradle:
...
+    cabal:
+      componentsToLoad: ["lib:foo"]
+      components:
+        - {"path": "./src", "component": "lib:foo"}
+        - {"path": "./app", "component": "exe:foo"}
+```
+
+The elements of `componentsToLoad` do not need to be also listed as individual components, though doing so might improve dependency tracking.
```
fendor, line 574: Will we use the `path` component of `components` to load the session even if `componentsToLoad` is given?

Saizan: we will use it to find the .cabal files to include in the dynamic dependencies, at least for cabal cradle

fendor: So we use both if both are given?

Saizan: Suppose you are working with a multi-package project, and this cradle:
```
cradle:
  cabal:
    componentsToLoad: ["lib:foo"]
    components:
      - {"path": "./foo", "component": "lib:foo"}
      - {"path": "./bar", "component": "lib:bar"}
```
then LoadUnitsFromCradle will load `lib:foo` and find a dependency on `./foo/foo.cabal`. Without the `components` field we would load the same but won't find the `foo.cabal`.

fendor: thank you for the explanation!

PR #516, opened 2026-06-29
Hide LoadMode in TestM

### LINKED ISSUE #509: `testLogger` should indicate which test the output id from
a possibility is to integrate biosTestCase* and runTestEnv so the latter can create a logger that prefixes the output with the TestName

fendor:
Small refactoring that fixes #509 by using `testCaseSteps`, which ensures the test logs are attributed to the correct test.

Further, we introduce more assertion helpers and hide the load mode in `TestM` allowing us to focus on the test definition instead of passing `LoadMode` everywhere.

The tests are now easier to analyse in CI, which is nice.

### DIFF
--- a/tests/BiosTests.hs
+++ b/tests/BiosTests.hs
@@ -4,7 +4,7 @@
 {-# LANGUAGE CPP #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE TypeApplications #-}
-module Main where
+module Main (main) where
 
 import Utils
 
@@ -92,7 +92,7 @@ symbolicLinkTests =
       assertCradle isMultiCradle
       step "Attempt to load symlinked module A"
       do
-        loadComponentOptions "./a/A.hs"
+        loadComponentOptions "./a/A.hs" []
         assertComponentOptions $ \opts ->
           componentOptions opts `shouldMatchList` ["a"] <> argDynamic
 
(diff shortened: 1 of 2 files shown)


soulomoon (approved):
LGTM

tests/Utils.hs
```diff
@@ -43,11 +45,10 @@ module Utils (
   initCradle,
   initImplicitCradle,
   loadComponentOptions,
-  loadComponentOptionsWithMode,
+  -- loadComponentOptionsWithMode,
```
fendor:
```suggestion
```

tests/Utils.hs
```diff
@@ -66,7 +67,7 @@ module Utils (
 
   -- * High-level test helpers
   testDirectoryM,
-  testDirectoryWithModeM,
+  -- testDirectoryWithModeM,
```
fendor:
```suggestion
```

tests/BiosTests.hs
```diff
@@ -436,7 +437,12 @@ hasOption mode flag munit opts = elem flag <$> do
               ]
             then pure unit_flags
             else pure []
-    options = componentOptions opts
+  liftIO $ case mode of
+    LoadFile -> return options
+    LoadFileWithContext -> expandGhcOptionResponseFile options
+    LoadUnitsFromCradle -> common
+    LoadUnitsInferred -> common
+  where
```
fendor:
```suggestion
```

PR #520, opened 2026-07-11
Make `cacheFile` thread-safe using file locks

crtschin:
This was mentioned on on Matrix. 

Changes include:

- Adds a `cacheFileIn` that allows varying the cache directory used, overriding the default behavior of using `$XDG_CACHE_HOME`.
- Made the above function thread-safe by using file-locks, and atomic file writes.

As long as downstream consumers were using the `srcHash` parameter as a content addressed tag, I don't think they have to necessarily use `cacheFileIn` and can keep using `cacheFile` as is after this change. I exported `cacheFileIn` mainly for the test I added.

### DIFF
--- a/src/HIE/Bios/Process.hs
+++ b/src/HIE/Bios/Process.hs
@@ -162,8 +162,8 @@ cacheFile fpName srcHash populate = do
 
 -- | 'cacheFile' with the cache directory given explicitly instead of resolved
 -- from the environment.
-cacheFileIn :: FilePath -> FilePath -> String -> (FilePath -> IO ()) -> IO FilePath
-cacheFileIn cacheDir fpName srcHash populate = do
+cacheFileIn :: CacheDir -> FilePath -> String -> (FilePath -> IO ()) -> IO FilePath
+cacheFileIn (CacheDir cacheDir) fpName srcHash populate = do
   createDirectoryIfMissing True cacheDir
   let newFpName = cacheDir </> (dropExtensions fpName <> "-" <> srcHash) <.> takeExtensions fpName
   -- Concurrent loads race to create the same cache entry, so serialize them on
--- a/src/HIE/Bios/Process.hs
+++ b/src/HIE/Bios/Process.hs
@@ -11,6 +11,7 @@ module HIE.Bios.Process
   , getCleanEnvironment
   -- * File Caching
   , cacheFile
+  , cacheFileIn
   -- * Find file utilities
   , findFileUpwards
   , findFileUpwardsPredicate
@@ -36,6 +37,7 @@ import qualified Data.HashMap.Strict as Map
 import Data.Maybe (fromMaybe)
 import qualified Data.Text as T
 import System.Environment
+import System.FileLock
 import System.FilePath
 import System.IO (hClose, hGetContents, hSetBuffering, BufferMode(LineBuffering), withFile, IOMode(..))
 import System.IO.Error (isPermissionError)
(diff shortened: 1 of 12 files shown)


fendor:
I think `hie-bios` caches need to be created from scratch for the tests in HLS. Otherwise, we risk the tests using existing caches (e.g., on a rerun), changing the test run. But we want the tests to be run the same way every time!

While the lock might be useful regardless, e.g. on windows this might be a thing that helps with multiple processes trying to open a file, I think we need to be able to specify the `XDG_CACHE_HOME` without having to read the global `XDG_CACHE_HOME` variable. I.e., I think we should add a parameter to `runCradle`, such as `CradleRunConfig` (completely made up name), which allows us to optionally set the cache root directory.

crtschin:
> Otherwise, we risk the tests using existing caches (e.g., on a rerun), changing the test run. But we want the tests to be run the same way every time!

Fair! Do the last two commits correspond to what you had in mind? I kept the previous API-exposed function that default to the environment usage. I can drop those if you don't think they're useful to keep.

src/HIE/Bios/Process.hs
```diff
@@ -156,14 +158,24 @@ readProcessWithOutputs outputNames l workDir cp = flip runContT return $ do
...
+-- from the environment.
+cacheFileIn :: FilePath -> FilePath -> String -> (FilePath -> IO ()) -> IO FilePath
+cacheFileIn cacheDir fpName srcHash populate = do
   createDirectoryIfMissing True cacheDir
   let newFpName = cacheDir </> (dropExtensions fpName <> "-" <> srcHash) <.> takeExtensions fpName
-  unlessM (doesFileExist newFpName) $ do
-    populate newFpName
-    setMode newFpName
+  -- Concurrent loads race to create the same cache entry, so serialize them on
+  -- a lock. Populate into a temp file and rename to keep it atomic as well.
```
fendor, line 170: This should only happen for the HLS tests because multiple processes are trying to write to the same directory, occasionally? HLS itself makes sure only one thread can run `runCradle`.

fendor (commented):
Yes, this is what I had in mind!

I think this is what we want. All the `FilePath`s are making me dizzy though, perhaps a `newtype CacheDir = CacheDir FilePath` is appropriate?

src/HIE/Bios/Process.hs
crtschin: I think this could also occur when multiple HLS instances are started fresh in the same workspace, on cold starts. But indeed that's a relatively infrequent, and probably unlikely occurrence on an unconventional dev workflow.

fendor: > I think this could also occur when multiple HLS instances are started fresh in the same workspace, on cold starts.

That should not happen and is a violation of our assumptions. `vscode-haskell` tries to make sure this doesn't happen and when it happens it is a bug.

crtschin: > vscode-haskell tries to make sure this doesn't happen and when it happens it is a bug.

Ensuring assumptions hold via client-side constraints is a bit fragile, I don't use HLS via vscode, for example. My masochistic tool of choice is helix. 

Just thought of this, but it's also not limited to HLS. This would be relevant for anyone relying on hie-bios that isn't overriding the cache directory with the new `CradleRunConfig`. If we don't go for this fix, we'll probably want to document that users don't have multiple pieces of software that rely on hie-bios starting up concurrently in their setups.

tests/BiosTests.hs
```diff
@@ -211,7 +241,9 @@ cabalTestCases extraGhcDep =
       initCradle "B.hs"
       assertCradle isCabalCradle
       root <- askRoot
-      buildDir <- liftIO $ cabalBuildDir root
+      buildDir <- liftIO $ do
+        cacheDir <- (fmap CacheDir . makeAbsolute . unCacheDir) =<< getCacheDir ""
```
fendor: I think it would be nice if there was something like `defCacheDir :: IO CacheDir`, this looks quite unwieldy.

fendor (approved):
LGTM, just one improvement to the testsuite, then this is ready to merge :)

tests/BiosTests.hs
crtschin: Agreed! Swapped this for `resolveCacheDir "" Nothing`.

PR #521, opened 2026-07-13
Add a CI timeout on tests

crtschin:
Noticed while working on #520, tests sometimes hang. This PR adds a step timeout to avoid jobs unnecessarily running for 6 hours.

### DIFF
--- a/tests/BiosTests.hs
+++ b/tests/BiosTests.hs
@@ -18,7 +18,7 @@ import HIE.Bios
 import HIE.Bios.Cradle
 import HIE.Bios.Cradle.Cabal (cabalBuildDir)
 import HIE.Bios.Types (LoadMode(..))
-import Control.Monad (forM_, forM, unless)
+import Control.Monad (forM_, forM, unless, when)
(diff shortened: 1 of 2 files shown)


tests/BiosTests.hs
```diff
@@ -69,6 +69,10 @@ main = do
   cabalDep <- checkToolIsAvailable "cabal"
   extraGhcDep <- checkToolIsAvailable extraGhc
 
+  -- Pre-create the shared extraGhc cabal store. The parallel tests otherwise
+  -- race to initialise it and can fail with "package.db already exists".
+  when (toolExists cabalDep && toolExists extraGhcDep) warmupExtraGhcStore
```
fendor, line 74: That sounds like a cabal bug, should we open an issue?

fendor (approved):
LGTM, thank you!

tests/BiosTests.hs
crtschin: Issue created [here](https://github.com/haskell/cabal/issues/12111).

